Issue #1363, opened 2026-09-16
fix(storage): release the generation file handle before Windows cleanup


DecisionNerd:
## Problem

The Python RC lane for Windows fails its clean-install native contract during cleanup:

```
PermissionError: [WinError 32] The process cannot access the file because it is being
used by another process:
  'C:\...\Temp\tmpyvlpydg3\persistent\generations\<uuid>\graph'
PermissionError: [WinError 5] Access is denied: '...\generations\<uuid>\graph'
    self._rmtree(self.name, ignore_errors=self._ignore_cleanup_errors)
```

A file inside a published generation is still open when the temporary project directory is removed. POSIX permits unlinking an open file, so Linux and macOS never surface this; Windows refuses, and the failure appears only there.

That makes it a handle-lifetime question rather than a Windows quirk to work around: something retains a file handle past the point the caller believes the project is closed. This repository already has prior art for lock and handle lifetime defects that only appear under specific conditions, in #275 and #469.

## Requirements

- Identify what holds the handle. Prefer closing it at the right point over suppressing the cleanup error.
- Do not resolve this with `ignore_errors=True` or a retry loop unless the retained handle is proven to be outside GraphForge's control, in which case record that proof.
- Existing crash-recovery, reopen and active-snapshot behaviour is unchanged. A handle released too early is worse than one released too late.
- The Windows native lanes remain real runtime authority; do not weaken them to pass.

## Acceptance criteria

- [ ] The Python RC lane for Windows completes its clean-install native contract.
- [ ] The retained handle is identified and named in the fix, not just made to stop erroring.
- [ ] A regression test covers the close-then-remove sequence on Windows, or the reason one cannot exist is recorded.
- [ ] Durability, recovery and reopen tests still pass on all platforms.

## Relationships

Child of #1359. One of four independent Binding RC platform failures. Related in class to #275 and #469.

Issue #1362, opened 2026-09-16
fix(ci): restore pnpm setup on Intel macOS for the Node RC lane


DecisionNerd:
## Problem

The Node RC lane for `x86_64-apple-darwin` fails during the pnpm setup action, before any GraphForge code runs:

```
npm error path /Users/runner/setup-pnpm/node_modules/@pnpm/exe
npm error command sh -c node setup.js
npm error @pnpm/exe does not ship a working binary for Intel macOS (darwin-x64)
          due to an upstream Node.js SEA bug.
```

`@pnpm/exe` has no working Intel macOS binary, so `pnpm/action-setup` cannot install on that runner. This is upstream, not a GraphForge defect, but it blocks a required publish lane.

`pnpm/action-setup` was bumped from 6.0.10 to 6.1.0 in #1264 on 2026-09-15, so a pin-back is the first thing to test. Be aware that the Binding RC failures predate that bump, so the version change cannot be the whole explanation and a pin alone may not fix it.

## Requirements

- The Intel macOS Node lane installs pnpm and completes.
- Prefer a fix that does not abandon Intel macOS support, since it is a declared publish target.
- If the resolution is to pin an action version, record which version works and why, so the pin is revisited rather than forgotten.
- If Intel macOS genuinely cannot be supported, that is a product decision about publish targets and must be raised rather than decided in CI configuration.

## Acceptance criteria

- [ ] The Node RC lane for `x86_64-apple-darwin` completes and produces its artifact.
- [ ] The chosen remedy is recorded with evidence, including which pnpm setup path works on `darwin-x64`.
- [ ] Other lanes using the same action still pass, including macOS ARM and Windows.
- [ ] No publish target is silently dropped.

## Relationships

Child of #1359. One of four independent Binding RC platform failures. Possibly interacts with #1264.

DecisionNerd:
Fix in #1367 (CI green, not merged).

[...]

Issue #1360, opened 2026-09-16
fix(bindings): make the Node package importable as an ES module


DecisionNerd:
## Problem

The Node RC lane for `x86_64-unknown-linux-gnu` fails its native parity and smoke contract:

```
SyntaxError: Named export 'GraphForge' not found. The requested module '../index.js'
is a CommonJS module, which may not support all module.exports as named exports.
not ok 1 - crates/graphforge-bindings-node/tests/smoke.test.mjs
```

`tests/smoke.test.mjs` is an ES module using a named import against the napi-generated `index.js`, which is CommonJS. Node cannot statically resolve named exports from CommonJS, so the import fails before any assertion runs.

This is a genuine packaging defect rather than infrastructure. A consumer doing `import { GraphForge } from '@curatelabs/graphforge'` from an ES module hits exactly this. The smoke contract exists to catch that, and it did.

## Requirements

- The published package is importable from an ES module using the documented import form, and from CommonJS, without the consumer knowing which module system the binding uses.
- Fix the packaging or the export surface, not the test. Weakening the smoke test to a default import would hide the consumer-facing problem it just caught.
- The generated TypeScript surface continues to describe the real exports.
- No change to class names, method names, or runtime behaviour.

## Acceptance criteria

- [ ] The Node RC lane for `x86_64-unknown-linux-gnu` passes its native parity and smoke contract.
- [ ] A named ES module import and a CommonJS require both resolve against the built package, proven by test rather than inspection.
- [ ] The generated `index.d.ts` matches the exports actually available at runtime.
- [ ] Other Node lanes that currently pass, Windows and macOS ARM, still pass.

## Relationships

Child of #1359. One of four independent Binding RC platform failures.

Issue #1340, opened 2026-09-16
ci(bazel): stop linking with the deprecated gold linker


DecisionNerd:
## Problem

Every Rust binary in the Bazel build links with the deprecated gold linker, and rustc warns about it once per link action.

```
warning: the gold linker is deprecated and has known bugs with Rust
  |
  = help: consider using LLD or ld from GNU binutils instead
```

In a successful Bazel Bootstrap job ([job 104931206646](https://github.com/CurateLabs/graphforge/actions/runs/35136770364)) this accounts for **123 of 256 build warnings**, one per linked Rust binary:

| Warning | Count |
| --- | --- |
| gold linker is deprecated | 123 |
| unreachable pattern | 2 |
| unused import `PathBuf` | 1 |
| function is never used | 1 |

The four warnings a reviewer should act on are buried under a repeated deprecation notice that nobody can act on from the log.

The claim is also not purely cosmetic. rustc states gold has known bugs with Rust, and binutils has deprecated gold for removal. This is a small correctness signal, not only noise.

## Where it comes from

Nothing in this repository selects a linker. `.bazelrc` sets no link options, and `.cargo/config.toml` only sets macOS `dynamic_lookup` flags. Bazel's auto-detected C++ toolchain adds `-fuse-ld=gold` when it finds gold on the machine, so the selection is implicit and environment-dependent.

That is why the warning never appears locally on a host without `ld.gold`, and always appears on the CI runners. The build's linker is currently decided by whatever happens to be installed on the executor.

## Why this is not a one-line change

The obvious fix is a single `--linkopt`. Two things make it larger than that.

- **Neither candidate is known to be present.** A development host checked for this issue has no `ld.gold` and no `lld`; its `ld` is GNU ld (bfd) 2.46. No workflow in this repository installs `lld`. Whichever linker is chosen must be verified to exist on the runners before it is configured, because a missing linker fails the build rather than falling back.
- **Release artifacts are built through Bazel and verified by hash.** Binding RC uses `native_builder: bazel` for Linux targets, and release certification compares retained bytes. Changing the linker changes the output bytes. That is acceptable, but it must land with the release lanes verified rather than as a side effect of an unrelated change.

## Objective

Select the linker explicitly rather than inheriting whatever the executor provides, remove the deprecation warning, and keep the release lanes provably intact.

## Requirements

[...]

PR #1365, opened 2026-09-16
fix(ci): repair the aarch64 Linux napi cross toolchain

### LINKED ISSUE #1361: fix(ci): repair the aarch64 Linux napi cross toolchain
## Problem

The Node RC lane for `aarch64-unknown-linux-gnu` fails while building its declared publish target:

```
Internal Error: Failed to set up the `--use-napi-cross` toolchain for aarch64-unknown-linux-gnu:
Cannot find native binding. npm has a bug related to optional dependencies
(https://github.com/npm/cli/issues/4828). Please try `npm i` again after removing both
package-lock.json and node_modules directory.
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command failed with exit code 1:
napi build --platform --release --target aarch64-unknown-linux-gnu --use-napi-cross
```

The cross toolchain cannot locate its own native binding. The cited npm optional-dependency bug is a known class, and the error text itself offers `--cross-compile` / `--use-cross` as alternatives to `--use-napi-cross`.

## Requirements

- The ARM64 Linux Node artifact builds reproducibly in CI.
- Prefer fixing the toolchain resolution over switching strategies blindly; if a different cross strategy is adopted, record why and confirm the produced artifact is equivalent in architecture and linkage.
- The artifact remains a real cross-built `aarch64` binary. Do not satisfy the lane by skipping the target or emitting a host-architecture binary.
- Note that #1340 changed the Linux linker to `ld.bfd`; if the cross toolchain interacts with that, record it.

## Acceptance criteria

- [ ] The Node RC lane for `aarch64-unknown-linux-gnu` builds and passes.
- [ ] The produced binding is verifiably `aarch64`, checked rather than assumed.
- [ ] The build is reproducible across two runs rather than passing once.
- [ ] Other Node lanes still pass.

## Relationships

Child of #1359. One of four independent Binding RC platform failures.

DecisionNerd:
## Problem

`Node RC (aarch64-unknown-linux-gnu)` has failed on every Binding RC run for days at
**Build declared publish target**:

```
Internal Error: Failed to set up the `--use-napi-cross` toolchain for aarch64-unknown-linux-gnu:
Cannot find native binding. npm has a bug related to optional dependencies
(https://github.com/npm/cli/issues/4828). …
```

## Root cause

`napi build --use-napi-cross` sets its toolchain up by downloading
`@napi-rs/cross-toolchain-x64-target-aarch64` and decompressing the bundled
`aarch64-unknown-linux-gnu.tar.xz` with `@napi-rs/lzma`
(`@napi-rs/cross-toolchain@1.0.3/index.js`, `download()`).

`@napi-rs/lzma` 1.5.0 raised its engine floor: 1.4.5 declares `engines.node: ">= 10"`,
while 1.5.0 and 1.5.1 — and **every one of their platform bindings** — declare
`engines.node: "^22.20 || ^24.12 || >=25"`.

pnpm silently skips an optional dependency whose engine range the running Node does not
satisfy. The lane ran Node 20, so `@napi-rs/lzma-linux-x64-gnu` was never installed,
`require('@napi-rs/lzma')` threw the generic napi loader message, and the CLI re-wrapped
it as the toolchain setup error. The npm optional-dependency bug the message cites is a
red herring — nothing is corrupt, the binding was correctly *not* installed.

Reproduced locally at the exact package count seen on the runner:

| Build-host Node | `pnpm install` | `@napi-rs/lzma-linux-x64-gnu` | `--use-napi-cross` |
| --- | --- | --- | --- |
| 20.19.5 | `Packages: +484` | skipped | fails, identical message |
| 22.22.1 | `Packages: +485` | installed | toolchain resolves, build succeeds |

The failing run logs `Packages: +484` on `v20.20.0`, matching the Node 20 row exactly.

This is a pure dependency-drift break, which is why it started without any workflow change.

## Fix

`node-version` for the Node RC job now comes from the matrix
(`${{ matrix.node_version || '20' }}`), and only the `aarch64-unknown-linux-gnu` lane
sets `node_version: "22"`.

- **The cross strategy is unchanged.** This repairs toolchain resolution rather than
  switching to `--cross-compile` / `--use-cross`, as the issue asked.
- **No lane is skipped and no host-architecture binary is emitted.**
- **The `engines.node: >=20` floor is not weakened.** This lane is
  `execution_mode: package-validation` — the x64 runner never executes the arm64 addon it
  builds — so its Node is a build host only. Every lane that actually runs the addon and
  the parity/smoke contract stays on Node 20.

`scripts/ci/test-binding-release-candidate.py` gains an assertion that the ARM64 cross
lane pins an explicit `node_version` of at least 22, that no other Node lane sets one, and
that the step sources the value from the matrix. Verified it fails when the lane is put
back on Node 20.

## Proof the artifact is genuinely aarch64

Ran the lane's exact command locally on an x86_64 host under Node 22:

```
pnpm --filter @curatelabs/graphforge exec napi build --platform --release \
  --target aarch64-unknown-linux-gnu --use-napi-cross
```

with `CFLAGS_aarch64_unknown_linux_gnu=-D__ARM_ARCH=8`, as the workflow sets. It produced
`crates/graphforge-bindings-node/graphforge.linux-arm64-gnu.node`:

```
$ file graphforge.linux-arm64-gnu.node
graphforge.linux-arm64-gnu.node: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV),
dynamically linked, not stripped

$ readelf -h graphforge.linux-arm64-gnu.node | grep -E 'Class|Machine|Type'
  Class:   ELF64
  Type:    DYN (Shared object file)
  Machine: AArch64

$ readelf -d graphforge.linux-arm64-gnu.node | grep NEEDED
  libdl.so.2  libgcc_s.so.1  libpthread.so.0  libm.so.6  libc.so.6

$ readelf -V graphforge.linux-arm64-gnu.node | grep -o 'GLIBC_[0-9.]*' | sort -u
  … GLIBC_2.17
```

`Machine: AArch64` rules out a host-architecture binary, the linkage is ordinary dynamic
glibc, and the highest referenced symbol version is `GLIBC_2.17` — the glibc baseline
`@napi-rs/cross-toolchain` documents, so the linkage is what the cross toolchain is
supposed to produce and not an accident of the build host.

The debug trace confirms the toolchain, not a host compiler, did the work:

```
napi:cross-toolchain Downloading @napi-rs/cross-toolchain-x64-target-aarch64@1.0.3 via npm pack ...
napi:build CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=~/.napi-rs/cross-toolchain/1.0.3/
             aarch64-unknown-linux-gnu/bin/aarch64-unknown-linux-gnu-gcc
```

## A second defect this uncovered

With the toolchain repaired, the lane reached `napi artifacts` for the first time and
failed there:

```
Internal Error: Missing artifacts for configured targets: aarch64-apple-darwin
(graphforge.darwin-arm64.node), x86_64-apple-darwin (…), x86_64-unknown-linux-gnu (…),
x86_64-pc-windows-msvc (…)
```

`collectArtifacts` reads `napi.targets` from the manifest and refuses to run unless an
addon exists for **every** one. This lane builds exactly one target — it is
`execution_mode: package-validation` precisely because the x64 runner cannot execute the
arm64 result — so four of the five are legitimately absent. The strict check is present in
`@napi-rs/cli` 3.8.6 through 3.9.1; the step had simply never been reached, because the
build died first.

The lane now passes `--config-path` with a config generated from its matrix target. The
release-candidate job, which downloads all five addons, still runs the same command against
the manifest, so the published package set remains gated on the full target list. The
contract test asserts that asymmetry in both directions.

Reproduced and fixed locally against the real cross-built addon: without the narrowed
config `napi artifacts` fails with the message above; with it the addon lands at
`npm/linux-arm64-gnu/graphforge.linux-arm64-gnu.node` and `validate-napi-artifacts.py`
accepts the result.

## Interaction with #1340

None. #1340 selected `ld.bfd` through `.bazelrc` `--linkopt`/`--host_linkopt`. This lane is
napi/pnpm/Cargo and never reads `.bazelrc`; `.cargo/config.toml` carries no Linux linker
override, and `--use-napi-cross` supplies its own
`CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER` (the toolchain's `aarch64-unknown-linux-gnu-gcc`,
which uses its own bundled `ld`). Recorded here because the issue asked.

## Verification

- Failure reproduced locally under Node 20 with the lane's exact command, then fixed under Node 22.
- Produced addon checked with `file` / `readelf` (above) rather than assumed.
- Binding RC dispatched twice from this branch against `main`'s SHA so the runner builds
  unmodified `main` code through the patched lane. Run ids recorded in a comment below.
- Sibling lanes that fail on those runs are #1360, #1362 and #1363, unchanged by this PR.

Closes #1361


DecisionNerd:
## Runner verification

Binding RC dispatched from this branch against `main`'s SHA `e378dd4d`, so the runner
builds unmodified `main` code through the patched lane.

| Run | Head of branch | `Node RC (aarch64-unknown-linux-gnu)` |
| --- | --- | --- |
| [35159879350](https://github.com/CurateLabs/graphforge/actions/runs/35159879350) | `f991c4be` (toolchain fix only) | build **succeeded**, failed later at `napi artifacts` — the second defect |
| [35160977991](https://github.com/CurateLabs/graphforge/actions/runs/35160977991) | `4681d0ba` | **success** |
| [35162039540](https://github.com/CurateLabs/graphforge/actions/runs/35162039540) | `4681d0ba` | **success** |

Two consecutive green runs at the same tree, so the lane is reproducible rather than
passing once.

Runner-side confirmation of the root cause, from run 35160977991's job log:

```
node-version: 22
Packages: +485
```

against `v20.20.0` / `Packages: +484` on the original failing run
[35155182821](https://github.com/CurateLabs/graphforge/actions/runs/35155182821). The one
extra package is `@napi-rs/lzma-linux-x64-gnu`.

### The artifact CI actually produced

Downloaded `binding-rc-addon-…-aarch64-unknown-linux-gnu` from both green runs:

```
$ file graphforge.linux-arm64-gnu.node
graphforge.linux-arm64-gnu.node: ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV),
dynamically linked, not stripped

$ readelf -h … | grep -E 'Class|Machine|Type:'
  Class:   ELF64
  Type:    DYN (Shared object file)
  Machine: AArch64

$ readelf -d … | grep NEEDED
  libdl.so.2  libgcc_s.so.1  libpthread.so.0  libm.so.6  libc.so.6

$ readelf -V … | grep -o 'GLIBC_[0-9.]*' | sort -u | tail -1
  GLIBC_2.17
```

`sha256` is identical across runs 35160977991 and 35162039540
(`c835c5c08633121de5194ddf2a273e81f79c69543ec1adba0af2064e23f08f3b`), and identical in
architecture and linkage to the addon built locally, so the artifact is byte-reproducible
as well as genuinely cross-built.

### Other lanes

Unchanged by this PR. On run 35162039540 the remaining failures are exactly the three
sibling children of #1359: `Node RC (x86_64-apple-darwin)` (#1360),
`Node RC (x86_64-unknown-linux-gnu)` (#1362) and `Python RC (python-windows)` (#1363).
`Node RC (aarch64-apple-darwin)`, `Node RC (x86_64-pc-windows-msvc)`,
`Python RC (python-ubuntu)` and `Python RC (python-macos)` all pass.

Issue #1384, opened 2026-09-17
perf(storage): eliminate the redundant authentication passes that dominate ingest


DecisionNerd:
## Maintainer direction

**Attack this overhead aggressively.** Prioritize speed and usability. The target for corruption handling is proportionate effort against corruption that would actually degrade the user experience, not maximum cryptographic assurance.

The burden of proof is reversed from the usual. Every authentication pass must justify its cost by naming a failure a user would notice and that nothing else catches. A pass that cannot do that is removed, not kept by default.

This does not authorize weakening crash safety, silently accepting corrupt data, or reducing the collision resistance of content-addressed naming. Those outcomes stay intact. What is being attacked is how many times the same bytes are read and hashed.

## The evidence

Ingest is 58 to 67% of total lifecycle wall time, and authentication is almost all of its I/O. Measured at 67,108,864 edges on commit `710c6c64`.

Ingest application I/O, by sub-phase:

| Ingest sub-phase | Read GB | Write GB |
|---|---:|---:|
| shape consume reauthentication | 145.9 | 113.4 |
| recovery reauthentication | 111.8 | 0.0 |
| encode write postwrite authentication | 46.6 | 12.2 |
| append merge | 0.0 | 25.0 |
| content-addressed install | 2.8 | 2.8 |
| hydration verification | 3.0 | 0.2 |
| **total** | **310.1** | **153.6** |

Authentication phases are **99.1% of ingest reads**. The actual merge writes 25 GB.

Normalized against what is kept:

| Quantity | Bytes per edge |
|---|---:|
| Retained on disk | 265 |
| Read for authentication | 4,579 |

**17.3 times the final data size, read purely to authenticate it.** The ratio is flat across a sixteen-fold range, so this is a structural constant, not a scaling artifact. Six of the nine tracked storage I/O phases are authentication or verification.

## What comparable systems do

External research across the Arrow, Parquet and DataFusion ecosystem found no precedent for this pattern.

- **InfluxDB 3** puts a CRC32 on write-ahead-log files, one per file, and on catalog files. Its Parquet data files carry no checksum and no hash. Its catalog records only path, size, row count and time bounds. The `sha2` crate appears in its dependency graph only for API token hashing, never for data.
- **GreptimeDB** checksums manifest checkpoint files and its write-ahead log. Its Parquet tables get nothing, and it does not enable the Parquet crate's checksum feature at all.
[...]

Issue #1269, opened 2026-09-14
fix(storage): authenticate retained shaping bytes during recovery


DecisionNerd:
## Verified problem

While validating #1268, a same-inode, same-length mutation of `shaped-identities.run` after a returned shaping failure was accepted on reopen. The failing native regression is preserved in `/tmp/gf1268-candidate-tests.log`. Independent source review traced the cause to unchanged main code: `authenticate_shaped_output` calls the writer-receipt fast path, which returns a stored digest after only inode/link-count/length checks. Comparing stored digests does not authenticate current payload bytes. Incomplete cleanup used this path without a subsequent payload check. Completed public recovery also calls the existing supersession payload authenticator; that protection remains unchanged.

## Scope and acceptance

- Reproduce incomplete-shape corruption acceptance on current main; retain direct regression coverage for both incomplete and completed shape refusal, valid reopen/retry and replaced-inode/extra-link refusal.
- Authenticate actual retained payload bytes at recovery/consumption trust boundaries. Keep efficient immediate writer accounting distinct; do not add compatibility readers or generic writer machinery.
- Charge actual authentication read work and cache-release work accurately. Preserve exact identity, properties, cancellation, recovery and successor authority.
- Prove corruption refusal without publication or destructive cleanup, and exact successful reconstruction/public lifecycle using relevant existing native fixtures.
- Run changed-surface gates and exact-head required CI, squash merge and verify closure before #1268 proceeds.

This independent correctness blocker is a native child of #1194 and blocks #1268. It does not authorize S24/S26 or a new certification workflow. #1194 remains the capacity close gate.

PR #1412, opened 2026-09-17
fix(storage): refuse completed-shape payload corruption at the boundary

### LINKED ISSUE #1392: fix(storage): completed-shape corruption is refused only incidentally, by a pass we are about to remove
## Problem

Same-inode, same-length corruption of a **completed** shape output is accepted. The refusal that ought to catch it exists only incidentally, as a side effect of a performance-motivated pass that is scheduled for removal.

This is the same class of defect as #1269, which was filed because a same-inode, same-length mutation of `shaped-identities.run` after a returned shaping failure was accepted on reopen.

## The gap

#1269's regression test loops over `completed ∈ {false, true}`.

- The **`false`** arm, an incomplete shape, is served by the documented #1269 fix.
- The **`true`** arm, a completed shape, is not. `authenticate_shaped_output` takes the writer-receipt fast path, which returns a stored digest after checking only inode, link count and length. Comparing stored digests does not authenticate current payload bytes, which is the exact diagnosis #1269 recorded.

The test passes today because the mutation is caught later and incidentally: `reclaim_superseded_payloads` calls `retire_payload`, which calls `authenticate_payload`, which streams the payload and computes a full SHA-256 before unlinking it. That pass exists to prove the right bytes are being deleted, not to guard this boundary.

## Why this is urgent now

That incidental pass is the largest single line in #1384's removal target, roughly 111.8 GB of reads at 67 million edges. **Removing it naively opens a corruption hole that is currently closed by accident.** Anyone optimizing the write path without knowing this would delete the protection and the test would still pass, because the test is satisfied by the pass being removed.

## Requirements

- Establish the refusal at the boundary deliberately rather than relying on a retirement-time side effect.
- The check must catch a same-inode, same-length payload mutation of a completed shape output, which is precisely what the writer-receipt fast path cannot see.
- It must survive the #1384 redesign. The write-path design proposes an inline payload checksum computed by the pass that writes the bytes, which is both stronger and roughly seventy times cheaper than a full re-hash at retirement; that is one acceptable shape, not the only one.
- Given the accepted decision in #1384, the primitive here may be non-cryptographic: the threat is byte mutation rather than a forged digest, and the threat model excludes an active same-identity adversary.

## Acceptance criteria

- [ ] A regression test fails on same-inode, same-length corruption of a completed shape output, and is proven to fail by mutation rather than only observed to pass.
- [ ] That test still passes when the retirement-time full re-hash is removed, proving the refusal no longer depends on it.
- [ ] #1269's existing regression test continues to pass on both arms.
- [ ] The cost of the new check is measured and reported per edge.

## Non-goals

- Removing the retirement-time pass. That belongs to #1384 and must not happen before this lands.
- Changing the threat model.

## Relationships

Blocks #1384's removal of the retirement-time authentication. Same class as #1269, which fixed the incomplete-shape arm of the same test. Found during the write-path design under #1387.

DecisionNerd:
## What this changes

`authenticate_shaped_output` — the trust boundary that a completed shape output
crosses on replay and on session reopen — took the writer-receipt fast path and
compared inode, link count, length and a *stored* digest. None of those can see
a payload mutation that preserves inode and length, which is the exact diagnosis
#1269 recorded for the incomplete-shape arm.

#1269's regression test still failed the open on the completed arm, but for the
wrong reason: `reclaim_superseded_payloads` → `retire_payload` →
`authenticate_payload` streamed and fully SHA-256'd **every superseded payload
immediately before unlinking it**, and tripped over the mutation on its way to
deleting the file. That pass is the largest single line in #1384's removal
target. Deleting it naively would have opened a corruption hole while the test
stayed green, because the test was satisfied by the pass being deleted.

So: establish the refusal deliberately, prove it by mutation, then delete the
pass it used to depend on — in that order, in one change, so the proof and the
removal cannot drift apart.

## Part one — the refusal (#1392)

- `ArtifactReceipt` gains `xxh64`, an inline corruption checksum over the same
  payload, produced by the same `HashingWriter` pass that writes the bytes. No
  pass reads a payload back in order to compute it.
- `authenticate_shaped_output` now streams the retained payload once and refuses
  it unless length **and** checksum match the receipt.
  `authenticate_shaped_output_identity` keeps the previous metadata-only check
  for the deletion paths, where the bytes are never consumed again.
- The verification runs **once per process-open**. `recover_shape_intent`
  performs it at session open; the completed-shape replay inside
  `shape_canonical_inner` skips it when this process already did.
  `prepare_canonical_encoding` no longer opts out of authenticating a replayed
  shape — it opted out because the retirement pass covered it.

New module `corruption_checksum` implements XXH64 (seed 0) and records, beside
the code, the two assumptions that permit a non-cryptographic primitive here:
ADR 0013's threat model excludes an active same-identity adversary, and ext4 on
mdadm RAID 1 does not checksum file data. If either stops holding, the decision
reverts. The implementation is pinned to the reference C implementation by its
published vectors. Content-addressed naming is untouched and stays SHA-256.

The value is serialised as fixed-width hex rather than a decimal `u64`, so a
receipt's byte length does not depend on its payload's content — the evidence
accounting compares those lengths.

### Mutation proof

`completed_shape_boundary_refuses_same_inode_payload_corruption` names the
boundary directly instead of observing `open_with_mode`. Against unfixed code it
fails; see the PR discussion for the run. It also asserts, in passing, that the
identity-only check the boundary used to perform still *accepts* the corruption
— the gap, pinned in place.

## Part two — remove the pass (#1384, design step S3d)

`retire_payload` and the completed-shape pass inside
`reclaim_superseded_payloads` drop their full SHA-256 re-reads and keep
identity, link count and length.

The only failure a re-hash there uniquely catches is *a payload was mutated
after it was consumed but before it was deleted*, which cannot change any
already-produced output. Everything still consumed is authenticated at the
boundary that consumes it: shape outputs by `authenticate_shaped_output`,
encoded artifacts by `authenticate_inventory_payloads`, shaped sources by the
encoder's spool digest — all of which run before a single predecessor is
unlinked.

## Measured

Repeatable fixture, real facade, `GraphConstructionBudgets::default()`, two
sizes:

| | 65,536 edges | 262,144 edges |
|---|---:|---:|
| bytes read per edge, before | 1,878.1 | 1,998.0 |
| bytes read per edge, after | **1,133.6** | **1,253.6** |
| | −39.6% | −37.3% |
| CPU µs per edge, before | 119.0 | 128.2 |
| CPU µs per edge, after | **104.7** | **114.4** |
| | −12.1% | −10.7% |
| recovery reauthentication bytes, before | 53,712,534 | 216,236,929 |
| recovery reauthentication bytes, after | **4,918,631** | **21,103,274** |
| | −90.8% | −90.2% |

Peak RSS moved by under 0.3% in either direction (79,148 → 78,892 KiB and
83,916 → 84,100 KiB) — noise, as expected from a step that deletes read passes
and adds no buffer.

On a successful single-process ingest the new boundary check costs **zero
bytes**: it only runs when a process opens a session whose shape is complete and
whose encoding is not, which is a crash-recovery path. On that path it replaces
one SHA-256 pass over the shaped output set with one XXH64 pass over the same
bytes.

## One behavioural change, named

`canonical_encoder_uses_owned_spool_after_source_mutation` asserted that a
shaped source mutated *after* the pass that consumed it made encoding fail. It
now succeeds: the encoder's owned spool holds the correct bytes, and the mutated
file is only being deleted. This is precisely the failure class #1384 accepts by
name. A mutation that reaches the consuming read still fails closed, unchanged,
and the test now asserts that alongside — the encoded artifacts still
authenticate, and the mutated source is retired rather than consumed.

`supersession_corrupt_successors_and_replaced_predecessors_fail_closed` keeps
its `shape` arm fail-closed; only the message changes, from the incidental
`payload digest changed` to the deliberate `shape manifest output payload
changed`.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace -- -D warnings`
- `cargo test -p graphforge-storage` and `cargo test -p graphforge-api`
- `shaping_recovery_refuses_same_inode_payload_corruption` passes **both** arms
  with the retirement re-hash removed — that is the proof the refusal no longer
  depends on the deleted pass.
- Corruption, cancellation, crash-boundary and recovery tests otherwise
  unchanged.

## Deliberately not in this change

- `authenticate_inventory_payloads` (the encoded-successor check inside
  `reclaim_superseded_payloads`) still full-SHA-256s the encoded graph, roughly
  3.3x per lifecycle. It is the only pass proving the successor is intact before
  the shape inputs are destroyed, and demoting it needs a checksum in the encoded
  inventory format. **That is deliberately out of scope here and is filed as
  #1417**, sequenced behind this PR so its baseline is measured after this lands.
  The recovery-phase cost that remains after this change is that pass, not an
  oversight.
- The wall-clock value embedded in `shaped-runtime-catalog.parquet`, which makes
  shaped output bytes non-reproducible across sessions. Surfaced by this work
  (the receipt checksum is stored as fixed-width hex precisely so a receipt's
  byte length does not depend on its payload). **Deliberately out of scope and
  filed as #1416**; it is load-bearing as a session binding token at
  `recovery.rs:275`, so the naive fix would remove a crash-recovery refusal.
- `authenticate_artifact` + `validate_parquet_metadata` at shape time (S3a), the
  encoder's post-write re-read (S3c), and the parent-inventory memoisation
  (S3f). S3b is already effectively free: the post-shape inventory pass measured
  2,080 bytes at 65,536 edges because the writer receipts are already on disk.
- `register-parquet`'s source copy, which the maintainer has confirmed is
  deliberate durability policy.
- The ADR for the authentication regime, which the design places at S7.

Closes #1392
Part of #1384

---
<a href="https://app.blacksmith.sh/CurateLabs/codesmith/graphforge/pr/1412"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792209813&installation_model_id=20486&pr_number=1412&repository=CurateLabs%2Fgraphforge&return_to=https%3A%2F%2Fgithub.com%2FCurateLabs%2Fgraphforge%2Fpull%2F1412&signature=13313f1ec2c58ad2f4d1ec79c29f767ad3949a9c6d0670c993628919ca1dd574"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


DecisionNerd:
## Mutation proof for the new refusal

The new regression test was run against **unfixed code** — a clean worktree at
`origin/main` (`a962f9cc`) with nothing applied but the test itself, adapted to
main's signature (`authenticate_shaped_output` returns `Result<(), GfError>`
there, and `authenticate_shaped_output_identity` does not exist yet):

```
$ cargo test -p graphforge-storage --lib -- \
    completed_shape_boundary_refuses_same_inode_payload_corruption --nocapture

running 1 test

thread '...::completed_shape_boundary_refuses_same_inode_payload_corruption'
panicked at construction_lifecycle_tests.rs:851:74:
called `Result::unwrap_err()` on an `Ok` value: ()

test ...::completed_shape_boundary_refuses_same_inode_payload_corruption ... FAILED

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1130 filtered out
```

`Ok(())` is the whole point: on main, `authenticate_shaped_output` **accepts** a
completed shape output whose first byte has been flipped, with inode, link count
and length all preserved. The test asserts a refusal and there is none.

On this branch the same test passes, and the assertion it makes on the way
through — that `authenticate_shaped_output_identity` (the check the boundary
used to perform) still returns `Ok` on the corrupted payload — pins the gap in
place so it cannot silently reopen.

The system-level counterpart is #1269's own
`shaping_recovery_refuses_same_inode_payload_corruption`, which passes **both**
arms on this branch **with the retirement-time re-hash removed**. That is the
proof the refusal no longer depends on the pass this change deletes.

Issue #1393, opened 2026-09-17
perf(storage): attribute and reduce the transient peak and the write amplification behind it


DecisionNerd:
## Why this is a release gate

S26 admission requires the projected transient peak to fit the host's free capacity with a 15% reserve. Today:

| Quantity | |
|---|---:|
| Projected S26 transient peak | 588.5 GB |
| Admission reserve | 141.3 GB |
| Available on the bench host | 739.3 GB |
| **Margin** | **9.6 GB, 1.6% of peak** |

A 1.6% margin on a projection is not a margin. And the two storage improvements that landed recently left the S22 transient peak **identical to the byte** across both ladder runs, so nothing so far has moved it. A 20% reduction turns 9.6 GB into 127 GB, which is the difference between hoping and knowing.

## Nobody knows what the peak is made of

Measured at 67,108,864 edges on `fa6447cc`:

| Quantity | |
|---|---:|
| Transient peak | 36.76 GB |
| Retained | 17.81 GB |
| Transient above retained | **18.95 GB** |

The existing `storage_attribution` categories account for only **2.79 GB** between them, so they explain neither the retained figure nor the peak. There is no breakdown of what occupies disk at the moment of peak.

**So this issue starts with measurement, not reduction.** This is the same situation as #1389, where reasoning from code structure to a conclusion about cost was wrong twice before measurement settled it.

## What is already known

- **The source copy in `register-parquet` cannot be removed.** It is deliberate durability policy: the session owns its bytes from registration so a user editing or deleting the source cannot affect an in-flight import. Hardlinking does not preserve that, since a shared inode makes an in-place edit visible. Reflink would, but the host is ext4 where it is unsupported, verified directly. On this filesystem a real copy is the only mechanism. **That portion of the peak is immovable.**
- **The write-path redesign may already reduce the peak.** Its range-partitioning removes the external merge tree entirely, which should remove merge intermediates from the peak. How much is unknown and should be measured rather than assumed.
- Write *volume* and transient *peak* are different quantities and only the peak gates admission. Physical writes are 363.8 GB against 17.81 GB retained, but that is bytes over time, not bytes resident simultaneously.

## Requirements

- **Attribute the transient peak.** What occupies the work root at the moment of peak, by category, and for how long. The existing category breakdown is inadequate and should be extended or replaced.
- Establish which portions are immovable, which are merge intermediates the redesign already removes, and which are genuine waste.
- Only then propose reductions, each with a measured expected saving.
[...]

PR #1413, opened 2026-09-17
fix(storage): promote same-filesystem CAS artifacts by rename, not copy

DecisionNerd:
Part of #1387

## What

The content-addressed install path copied every payload byte (read source, write temp, hash while copying) even when the source and the CAS temp directory live on the same filesystem. This adds `install_graph_object_file_by_move`, which promotes a same-filesystem source into the CAS by `renameat` + `linkat` instead: zero write bytes, one read pass to authenticate the staged bytes against their declared digest before sealing (unchanged: the temp must be sealed read-only before linking, `validate_sealed_temporary` still asserts link count 1, and the digest is verified from the installed inode, never trusted from the caller).

**Verified the design's claim that CAS publish already hardlinks** (`finalize_temporary_object`'s `cas.tmp.link_child_into` call, `installation.rs`) — true, unchanged by this PR. The remaining copy was specifically the workspace→tmp staging step, confirmed by reading and by measurement (below).

## Why this is scoped as an explicit opt-in, not a blanket change

The design's own §6.2 called this section a "thinner pass," and verifying it turned up two things a blanket rewrite of `install_graph_object_file_with_lease`/`append_graph_files_v2` would have broken or silently weakened:

1. **`append_graph_files_v2`'s own test suite installs the same workspace content more than once** — into a second, independent container, and by re-adding a previously tombstoned path from the same workspace after an earlier call already consumed it. A rename-based install destroys the source name, which broke `radix_root_is_deterministic_across_descriptor_order` and `patricia_delete_collapse_and_readd_converge_to_fresh_roots` outright when I first tried making the fast path the default. That's real, currently-relied-on behavior, not a defect in those tests.
2. **A rename shares the source's inode with the installed CAS object.** A descriptor opened on the source before install and still open afterward can mutate the sealed object's bytes; a byte copy is immune to this by construction (confirmed empirically: `chmod`-to-read-only does not revoke write access through an already-open file descriptor on Linux). This is a real, narrower guarantee than the copy path provides — analogous to, but distinct from, the already-settled register-parquet hardlink-vs-copy tradeoff — and is now documented and covered by `cas_move_shares_the_source_inode_with_a_preexisting_writable_descriptor`.

Given that, only `append_authenticated_mapped_graph_files` — the checkpoint-authorized mapped encoding path `graph_construction::encoding_publication` uses once per generation for its own freshly written, session-exclusive workspace — opts in, via a new `move_eligible` parameter threaded through `append_graph_files_v2_inner`. Every other append/install entry point (plain append, replay, mapped import, v1-to-v2 migration) is byte-for-byte unchanged, defaulting `move_eligible = false`.

## Crash recovery

A crash between the rename and the final `linkat` is recoverable: the staged file's name is derived from the digest (`moving-<digest>`) rather than a fresh random UUID, so a retry with the same source path and digest resumes the same staged bytes even though the original source name is gone by then. `cas_move_resumes_after_a_crash_between_the_move_and_the_final_link` proves this by injecting a fault at the `install:final-linked` boundary and retrying. All existing crash-boundary, corruption-refusal, and cancellation tests are unweakened (see Verification).

Register-parquet's source copy is untouched — settled, deliberate durability policy per the brief; this content-addressed CAS step never had to guarantee source immutability for its own construction-root artifacts the way register-parquet does for the user's file.

## Measured: bytes written per artifact, copy vs. move

Rerunnable fixture: 200 artifacts × 64 KiB through each path.

| Path | write_bytes | write_calls |
|---|---:|---:|
| Copy (unchanged, all other callers) | 13,107,200 (1.0 write byte / logical byte) | 200 |
| Move (`move_eligible = true`, encoding_publication's call site) | 0 | 0 |

## Verification

- `cargo fmt --all -- --check`: clean.
- `cargo clippy -p graphforge-storage --all-targets -- -D warnings`: zero new warnings from this change (confirmed by diffing clippy output against `origin/main` — both report the same 257 pre-existing errors in unrelated files, e.g. `writer/property_mutation/tests.rs`, `writer/replay_properties/tests.rs`; none in any file this PR touches).
- `cargo test -p graphforge-storage --lib graph_object_store::`: 45/45 pass, including 5 new tests (`cas_move_promotes_a_same_filesystem_source_by_rename_and_writes_nothing`, `cas_move_shares_the_source_inode_with_a_preexisting_writable_descriptor`, `cas_move_refuses_corruption_introduced_before_installation_authenticates_it`, `cas_move_falls_back_to_copy_when_ineligible`, `cas_move_resumes_after_a_crash_between_the_move_and_the_final_link`) and 3 that needed reworking after the scope narrowed from "blanket" to "opt-in" (`patricia_delete_collapse_and_readd_converge_to_fresh_roots`, `publication_components_measure_fresh_reuse_and_path_copy_work`, `radix_root_is_deterministic_across_descriptor_order` — all pass unmodified now that the plain append path stays copy-based).
- `cargo test -p graphforge-storage --lib` (full suite) and `cargo test -p graphforge-api --lib` (full suite): diffed the failing-test-name sets against an unmodified `origin/main` checkout run the same way. Every failure is identical on both branches (~278 in storage, ~245 in api), all `Project { code: UnsupportedFilesystem, message: "phase=CLASSIFY outcome=rejected cause=filesystem_class_unproven" }` — a pre-existing, environment-dependent `filesystem_admission` probe failure in this sandbox, unrelated to this change. Zero test names fail on this branch that don't also fail on `origin/main`.
- `graph_construction::encoding_publication::` tests specifically: 9 pass / 25 fail, identical set on both branches for the same reason above. I could not get a full green run of this module's real `append_authenticated_mapped_graph_files(move_eligible = true)` integration path through `cargo test` in this sandbox as a result — every test in that file that reaches full publication is blocked by the pre-existing environmental gate. I verified the wiring by code reading instead (see PR description above and inline comments), and confirmed the one check the design flagged as needing adjustment (`encoding_publication.rs`'s post-install `path_identity(&workspace) != encoded_directory.identity()` check) inspects the **workspace directory's** own inode, not the individual artifact files inside it, so it is unaffected by files being renamed out of that directory.
- Crash/corruption-specific: `install_crash_boundaries_leave_only_valid_or_recoverable_unreferenced_state`, `cas_move_resumes_after_a_crash_between_the_move_and_the_final_link`, `cas_move_refuses_corruption_introduced_before_installation_authenticates_it`, `canonical_encoder_cancellation_recovers_and_corruption_fails_closed` (encoding_publication) all pass.

## A process note for whoever picks up the next storage agent task

Mid-task, a `git stash`/`git stash pop` I ran (to diff my branch against `origin/main` without a second worktree) collided with another agent's concurrent `git stash` in a *different* worktree of this same clone — `refs/stash` is shared across all worktrees of one `.git` directory, it is **not** worktree-isolated the way the working tree and index are. My pop applied someone else's WIP (`fix/1403-1404-discarded-reopens`) into my working tree instead of my own changes, and their stash entry was dropped from the shared stash list. I traced both commits via `git fsck --unreachable`, restored their entry with `git stash store`, and recovered my own from its dangling commit — no data was lost, but it was close. Recommend adding this to the standing brief: never use `git stash` for a baseline comparison in a shared-clone worktree; use `git worktree add <path> <commit>` (a second, fully separate worktree) or a patch file instead.

---
<a href="https://app.blacksmith.sh/CurateLabs/codesmith/graphforge/pr/1413"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792210290&installation_model_id=20486&pr_number=1413&repository=CurateLabs%2Fgraphforge&return_to=https%3A%2F%2Fgithub.com%2FCurateLabs%2Fgraphforge%2Fpull%2F1413&signature=70a5c385a9ba46dc049e13fa3af6a038d968066ef5618e6952b3ea75f8bc6bf5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


DecisionNerd:
## Deliberately held pending a measurement, not stalled

This PR is **not** waiting on review or on its CI failures. It is held until a running measurement decides whether it is worth its tradeoff. Recording the reasoning so nobody merges or closes it uninformed.

### What changed since it was opened

Its original justification was the S26 disk gate. **That justification is gone.** #1393's transient-peak composition measured the S22 peak instant and found the content-addressed staging copy is **zero at the admission peak** — it is built after the merge collapses, and its own high-water mark sits *below* the merge tree's. So this change does not move the number that gates S26 admission at all.

Separately, S5 (#1430) deleted the external merge tree and the post-redesign peak has now been measured at **402.6 B/edge** against #1393's 400.3 projection — 0.6% agreement. **The disk gate is closed without this PR.**

### What remains, and why it is not obviously worth it

The write reduction is real: 13,107,200 bytes and 200 write calls become **0** for the one opt-in call site. But measured ingest reads **1,973 B/edge** and writes **713** — reads are 2.8x writes, and CPU per edge is what sets how many cores the throughput floor demands. This change is on the smaller axis.

Against that it narrows a durability guarantee: a rename **shares the source inode** with the installed CAS object, so a descriptor opened before install and still open afterward can mutate sealed bytes. A byte copy is immune by construction. That was found, documented and tested here rather than discovered later, which is to this PR's credit — but it is a real narrowing, correctly scoped to a single session-exclusive call site because making it the default broke two tests that legitimately install the same workspace content twice.

### What decides it

A measurement in flight is decomposing the **append + seal + publish** tail — **17.3% of ingest wall, parallelism unknown** — which is where CAS install lives.

- **If CAS install is a meaningful share of that 17.3%:** merge, fix the three failing Bazel targets, keep the narrow opt-in scope.
- **If it is a small share:** close. The inode tradeoff is not worth an unquantified gain on the wrong axis.

### Until then

**Do not fix the Bazel failures** (, , ). That work is wasted if the answer is close, and no agent has been assigned to it for that reason.

Note the ladder failure here is  — this PR making install writes O(1) is exactly what trips it. That is catalogued in #1433 as one of four rows where the redesign broke a scale-bearing assertion, and **two of those four turned out to be real defects rather than legitimate flattening**, so it should not be waved through on the assumption that flat means the redesign is working.

DecisionNerd:
*(Correcting the previous comment — several target names were mangled by shell expansion when it was posted. Full text below; the previous one can be ignored.)*

## Deliberately held pending a measurement, not stalled

This PR is **not** waiting on review or on its CI failures. It is held until a running measurement decides whether it is worth its tradeoff. Recording the reasoning so nobody merges or closes it uninformed.

### What changed since it was opened

Its original justification was the S26 disk gate. **That justification is gone.** #1393's transient-peak composition measured the S22 peak instant and found the content-addressed staging copy is **zero at the admission peak** — it is built after the merge collapses, and its own high-water mark sits *below* the merge tree's. This change does not move the number that gates S26 admission.

Separately, S5 (#1430) deleted the external merge tree, and the post-redesign peak has now been measured at **402.6 B/edge** against #1393's 400.3 projection — 0.6% agreement. **The disk gate is closed without this PR.**

### What remains, and why it is not obviously worth it

The write reduction is real: 13,107,200 bytes and 200 write calls become **0** for the one opt-in call site. But measured ingest reads **1,973 B/edge** and writes **713** — reads are 2.8x writes, and CPU per edge is what sets how many cores the throughput floor demands. This is on the smaller axis.

Against that it narrows a durability guarantee: a rename **shares the source inode** with the installed CAS object, so a descriptor opened before install and still open afterward can mutate sealed bytes. A byte copy is immune by construction. That was found, documented and tested here rather than discovered later, which is to this PR's credit — but it is a real narrowing, correctly scoped to a single session-exclusive call site because making it the default broke two tests that legitimately install the same workspace content twice.

### What decides it

A measurement in flight is decomposing the **append + seal + publish** tail — **17.3% of ingest wall, parallelism unknown** — which is where CAS install lives.

- **If CAS install is a meaningful share of that 17.3%:** merge, fix the failing Bazel targets, keep the narrow opt-in scope.
- **If it is a small share:** close. The inode tradeoff is not worth an unquantified gain on the wrong axis.

### Until then

**Do not fix the Bazel failures** — `scale_g500_ladder`, `graphforge_api_test`, `graphforge_storage_test`. That work is wasted if the answer is close, and no agent has been assigned to it for that reason.

Note the ladder failure here is `cas_install_read_write.write_bytes has no positive growth` — this PR making install writes O(1) is exactly what trips it. That is catalogued in #1433 as one of four rows where the redesign broke a scale-bearing assertion, and **two of those four turned out to be real defects rather than legitimate flattening**, so it must not be waved through on the assumption that flat means the redesign is working.

DecisionNerd:
## Closing: the measurement came back, and CAS install is 1.8%

This PR was held pending a decomposition of the append/seal/publish tail, with a decision rule recorded above: **merge if CAS install is a meaningful share, close if it is small.** The measurement is in.

Per-phase read bytes, measured on `main` at three scales via `ConstructionPhaseAttribution` (the per-phase I/O attribution from #1422):

| phase | B/edge at 8.4M edges | share of all reads |
|---|---:|---:|
| ShapeConsumeReauthentication | 964.0 | 40.0% |
| RecoveryReauthentication | 819.3 | 34.0% |
| EncodeWritePostwriteAuthentication | 349.3 | 14.5% |
| SealAuthentication | 236.6 | 9.8% |
| **CasInstallReadWrite** | **43.0** | **1.8%** |

And the coarse phase timing, same run:

| phase | 8.4M edges | share of wall | growth exponent |
|---|---:|---:|---:|
| publish_canonical (contains CAS install) | 3.097s | 3.1% | 0.96, linear |

So the phase this PR optimises is **1.8% of read bytes and ~3% of wall time, growing linearly**. Removing its copy entirely cannot move the throughput floor, which needs 10-20x.

## Why that settles it

The original justification — the S26 disk gate — was already gone: #1393 measured the content-addressed staging copy as **zero at the admission peak**, and S5 (#1430) has since closed the capacity gate outright, with the post-redesign peak measured at **402.6 B/edge** against a 400.3 projection.

What remained was a throughput argument. It is now sized, and it is small — on the *write* axis, where ingest writes 713-1043 B/edge against **2,412 B/edge read**. The levers that matter are elsewhere, and they are now measured:

- **RecoveryReauthentication, 34% of all reads.** `reclaim_superseded_payloads_cancellable` re-authenticates payloads *immediately before deleting them*, on the ordinary happy path, inside a process that already validated an unbroken identity chain for those bytes. Tracked as #1417.
- **ShapeConsumeReauthentication, 40% of all reads** and the only clearly superlinear read line (1.34x over a 16x span). Its dominant piece is the k-way merge tree's own multi-pass I/O — which **S5 has already deleted**.

Against a 1.8% gain, this PR narrows a real durability guarantee: a rename shares the source inode with the installed CAS object, so a descriptor opened before install and still open afterward can mutate sealed bytes. A byte copy is immune by construction.

**That trade is not worth making for 1.8%.**

## What was good here, and is preserved

The work stands on its own merits and is recoverable from this branch if the calculus ever changes — for instance if the read-side levers land and CAS install becomes a proportionally larger share.

Particularly worth keeping: this PR **found and documented the inode-sharing hazard itself**, with an empirical confirmation that `chmod` to read-only does not revoke write access through an already-open descriptor on Linux, and a test covering it. It also correctly narrowed scope to a single session-exclusive call site after discovering that two existing tests legitimately install the same workspace content twice. That is the right way to handle a subtle tradeoff — surface it, scope it, test it — and it is why the decision could be made on evidence rather than guesswork.

Closing as **not worth its tradeoff at the measured size**, not as wrong.

Part of #1387.

Issue #1388, opened 2026-09-17
epic(query): bounded queries should cost what their results cost, not what the graph costs


DecisionNerd:
## Goal

Make bounded queries cost what their results cost, not what the graph costs.

**Floor: a bounded query returns in under one second at one billion edges.** Today the equivalent query takes about eighty seconds at 67 million edges and scales linearly from there.

## The finding

Two canonical queries are run at every ladder rung: a one-hop and a two-hop traversal, each with `ORDER BY ... LIMIT`. Both are bounded by construction. The engine records exactly how much of the graph they touch.

Measured on `fa6447cc`, ladder run completed 2026-09-17:

| Rung | Live edges | Query phase | Adjacency rows examined |
|---|---:|---:|---|
| S18 | 4,194,304 | 11.0 s | 46 and 4 |
| S19 | 8,388,608 | 20.2 s | — |
| S20 | 16,777,216 | 42.9 s | 78 and 6 |
| S22 | 67,108,864 | 179.8 s | 122 and 6 |

**Two queries that examine 128 adjacency rows between them take 180 seconds.** Across a sixteen-fold increase in graph size the work performed grows by a factor of roughly two, while the time grows by a factor of sixteen point three.

Normalized, the cost is essentially constant per edge in the graph: 2.63, 2.41, 2.55 and 2.68 microseconds per edge at the four rungs. That is the signature of work proportional to the dataset, not to the query.

The same shape appears in every phase that opens a project:

| Rung | Reopen | Recount | Query | Reopen proof |
|---|---:|---:|---:|---:|
| S18 | 0.82 µs/edge | 0.52 | 2.63 | 3.69 |
| S19 | 0.66 | 0.41 | 2.41 | 3.58 |
| S20 | 0.80 | 0.50 | 2.55 | 3.63 |
| S22 | 0.78 | 0.49 | 2.68 | 3.66 |

Every column is flat. **Opening a project costs more per edge than counting every edge in it**, 0.78 microseconds against 0.49. That is not a plausible cost for an operation that should resolve a pointer and load an index.

A bounded query should cost roughly what its result costs, or at worst grow logarithmically with the graph. Ours grows linearly.

## Why this matters more than the wall-clock number

This is the same architectural shape as #1094, which found that opening the sharded adjacency index decoded every shard, making resident memory scale with edge count. That issue fixed the memory growth and the ladder now passes its bounded-memory gate. The **time** cost of touching the whole graph appears to remain.

[...]

Issue #1094, opened 2026-09-04
fix(exec): plateau S18→S19 recount/query RSS so host ladder admits S20


DecisionNerd:
## Problem

Canonical GraphForge lifecycle ladder on OVHC-AGENCY on `e57ccbb3` (after #1092/#1090) completed S18 and S19 under `local-linux-cgroups-v2`, then S20 failed closed with typed `projection_refused`.

Authoritative evidence under `/home/ubuntu/graphforge-ladder/evidence-official/` (commit-bound):

| Rung | live_edges | peak_rss_bytes | wall_s |
|---|---|---|---|
| S18 | 4,194,304 | 655,425,536 (~0.61 GiB) | 195 |
| S19 | 8,388,608 | 1,234,415,616 (~1.15 GiB) | 599 |

Adjacent RSS growth fraction ≈ **0.883** vs gate `max_adjacent_rss_growth_fraction = 0.1` (`rss_bounded_or_plateaued`).

## Remaining failure after #1124 (2026-09-06)

#1124 landed `EdgeCountExec` and ordered one-hop bounds. Re-run on `007edc1e` still refuses S20:

| Rung | live_edges | peak_rss_bytes | wall_s |
|---|---|---|---|
| S18 | 4,194,304 | 551,989,248 | 98 |
| S19 | 8,388,608 | 1,030,402,048 | 196 |

`rss_growth_fraction = 0.8667` vs 0.1. Preserve `/home/ubuntu/graphforge-ladder/1094-007edc1e/evidence/`.

Confirmed remaining cause: `ShardedCsrIndex::open` decoded every shard at open (O(E) RSS). Host `MATCH ()-[r]->() RETURN count(r)` then called `provider.adjacency()`. One-hop is already bounded (`candidates_generated=1000`, `adjacency_rows_examined=46`).

Repair: (1) open validates manifest + stats shard files only; checksum stays on first row touch; (2) `AdjacencyProvider::edge_cardinality` reads stamped manifest `edge_count` when the index is fresh and deltas are empty. Do not relax the 10% gate or change host queries to `-[r:LINK]->`.

## Objective

Make recount edge-count and canonical one-hop ordered-LIMIT remain bounded so adjacent S18/S19 process RSS plateaus within the progressive gate (≤10% growth) and S20 is admitted from host capacity on OVHC-AGENCY. No ladder-only path; no gate relaxation.

## Requirements

- Keep sharded CSR open off the O(E) decode path.
[...]

Issue #1417, opened 2026-09-17
perf(storage): add a checksum to the encoded inventory so reclaim stops re-hashing every payload byte


DecisionNerd:
## Summary

After #1392 lands, `authenticate_inventory_payloads` is the largest remaining cost in the recovery phase: a full SHA-256 over every byte of every encoded artifact, run inside `reclaim_superseded_payloads`. #1392 cut recovery reauthentication bytes by ~90% at two scales and deliberately stopped here, because demoting this one needs a change to the **encoded inventory format** rather than a change to call sites.

Filing so the remaining cost has an owner and is not read as an oversight in #1392's review.

## Where it is

    crates/graphforge-storage/src/graph_construction_encoding.rs:2650
        pub(crate) fn authenticate_inventory_payloads(...)

    crates/graphforge-storage/src/graph_construction/supersession.rs:59
        called from the supersession / reclaim path

    crates/graphforge-storage/src/graph_construction_encoding.rs:2613
        second call site

## Why it cannot be deleted the way the other passes could

The passes #1392 removed were **redundant** — something else already established the same property, so removing them cost nothing. This one is not redundant. It is the only thing standing between a corrupted encoded artifact and its adoption during reclaim. Deleting it would mean accepting bad data silently, which is out of bounds.

The settled policy for this codebase is: **hash once on write; cryptographic digests where a digest names an object; a fast checksum for corruption detection; full intactness checking behind an explicit command.** This call site is doing corruption detection with a cryptographic digest, which is the expensive combination and the wrong one for the job.

Note the standing project direction — *this is an analyst bench, not a forensic system* — authorises removing verification that only ever catches corruption in bytes nobody reads. It does **not** authorise accepting corrupt data, weakening crash recovery, or reducing collision resistance anywhere a digest is used as a content-addressed **name**. Check which role the digest plays at each call site before changing it; if it is a lookup key, it stays a SHA-256.

## The actual work

Add a checksum to the encoded inventory format so corruption can be detected without re-reading and re-hashing every payload byte:

1. Establish which of the two call sites need corruption detection and which (if either) need a content-addressed name. They may not be the same.
2. Add a fast checksum field to the encoded inventory entry. This is a durable-format change, so it needs a version bump and a read path for inventories written before it, and it should be stated in the design and recorded as an ADR afterwards, per the standing decision on the write-path redesign.
3. Demote the corruption-detection call site to the checksum. Leave any name-forming digest alone.
[...]

PR #1425, opened 2026-09-17
perf(storage): memoize and cheapen the open-path generation inventory sweep

DecisionNerd:
## Summary

Part of #1388. Implements steps O1, O2, and O3 of the open-path redesign (design doc reviewed at session start; O3 explicitly approved by the maintainer mid-implementation as an in-scope extension — see reasoning below). Does **not** implement O4 (XXH3 swap in `ordinal_identity_v4.rs`) or anything past O3 — those are owned by other agents / require separate sign-off.

`ResolvedProjectGeneration::graph_files_inventory()` decoded the manifest and, for the V2/mapped-root participant, ran a full streamed SHA-256 over every declared graph payload object. This function was called at least four times inside a single `GraphForge::open`, so a V2/mapped-root open paid for four-plus full-byte-range SHA-256 passes over the same graph payload.

- **O1** — memoize `graph_files_inventory()` on `ResolvedProjectGeneration` via `Arc<OnceLock<Result<Option<Arc<GraphFilesInventory>>, GfError>>>`. A resolved generation is immutable by construction (`CURRENT` is never re-consulted after resolution), so every call site, and every clone of a `ResolvedProjectGeneration` for the same generation, now shares one computation. Zero behavior change to what gets verified.
- **O3** — delete the payload `verify_graph_object` call inside `graph_files_inventory()`'s V2 loop; keep the existing presence-and-declared-length check. This is the step that removes the O(E) term. Every real consumer of a graph object's bytes already re-authenticates lazily on first touch via `open_graph_object_by_digest`/`read_graph_object_by_digest`, so the eager sweep was provably redundant for objects a session reads, and simply undetected-until-next-read for objects it does not. Mirrors the presence-only pattern `ShardedCsrIndex::open` already uses for adjacency shards (#1094).
- **O2** — replace `materialize_from_cas`'s post-hardlink re-verify (`graph_object_store/materialization.rs`) with an identity, length, and link-count check. `link_child_into` already confirms the newly linked directory entry resolves to the same inode as the CAS source; nothing between the link syscall and the check can have altered those bytes within the same open, so re-reading and re-hashing the whole object was pure duplicate work.

### The tradeoff O3 makes (maintainer-approved)

Before: `GraphForge::open` proved every declared graph object was byte-for-byte intact before returning — including objects no query in the session ever reads. After: open proves every declared object is present at its declared length; content authentication is deferred to first real read (which already happens unconditionally) or to the explicit administrative verify command (#1384 decision 4). A corrupted object that a session never reads is no longer caught at open — it surfaces on the next read of that object, or on an explicit verify run. Corruption in an object a query *does* read is still caught, exactly as before, because the lazy per-object authentication path is untouched. The maintainer's framing, recorded here per the coordinator's request: the verify command covers whole-store integrity, this path maximizes usable speed, and this is an analyst bench rather than a forensic system.

### One correction to the design doc

The design cited `GraphObjectReadLease::open_for_construction` (`graph_object_store.rs:959-1054`) as an existing precedent for "trust identity + length without re-hashing." Reading it directly: it still streams the whole file through SHA-256 and compares the digest — it does **not** skip hashing. The real precedent for O2's metadata-only check is `StableDirectory::link_child_into` itself (`graphforge-filesystem/src/lib.rs`), which already asserts `file_identity(installed) == file_identity(source)` as part of completing the hard link, plus the identity/link-count revalidation pattern already used elsewhere in `graph_object_store.rs` (e.g. `revalidate_lifecycle`, `replace_child`). O2 mirrors that actual pattern, not `open_for_construction`.

## Verified

All runs below compare this branch against `origin/main` at the commit this branch is based on, back to back, on the same host.

- `cargo fmt --all -- --check` — clean.
- `cargo clippy --workspace -- -D warnings` — clean, zero warnings, full workspace.
- **openCypher TCK corpus** (`cargo test -p graphforge-api --test bdd --release`): **3,897/3,897 scenarios passing, identical before and after** (verified via the harness's structured `report.json`, not just the summary line).
- **API BDD suite** (same binary, 118 scenarios): 115 passed / 3 failed, **identical failing scenario set before and after**. All 3 failures are pre-existing and environment-caused (`UnsupportedFilesystem`/`filesystem_class_unproven` — the CI/dev sandbox's `/tmp` is tmpfs, which `filesystem_admission.rs` correctly refuses to classify as durable storage), not caused by this change.
- **`graphforge-storage --lib`** (1,128 tests, clean rerun on ext4-backed worktrees after tmpfs was found to be the confound above): **850 passed / 278 failed, identical failing test set before and after**. Every sampled failure traces to the same `filesystem_class_unproven` cause in `filesystem_admission.rs`, a module this PR does not touch.
- **`graphforge-api --lib`** (742 tests — includes `durability_certification_tests`, `composite_recovery_tests`, `mutation_transaction_fault_tests`, `stream_cancellation_isolation_tests`, `workspace_hydration` tests, corruption/crash-recovery/cancellation coverage), run with `TMPDIR` pointed at an ext4-backed directory to remove the tmpfs confound entirely: **742 passed / 0 failed, both before and after this change.**

## Measured: open cost per edge, before vs. after

Two graph sizes (8x range), each opened repeatedly (10-15 iterations) from a durably-published generation, wall-clock measured around `GraphForge::new(Some(path))`, on the same ext4-backed host, back to back:

| Edges | Before: avg open | Before: µs/edge | After: avg open | After: µs/edge |
|---:|---:|---:|---:|---:|
| 50,000 | 108.2 ms | 2.16 | 110.3 ms | 2.21 |
| 400,000 | 1515.1 ms | 3.79 | 785.9 ms | 1.96 |

At 400,000 edges: **1.93x wall-clock speedup**, per-edge cost drops **48%** (3.79 → 1.96 µs/edge). The qualitative shape matches the design's prediction: *before*, per-edge cost grows with graph size (3.79 vs 2.16 µs/edge, 8x range) — the O(bytes) signature; *after*, per-edge cost is flat-to-decreasing (2.21 vs 1.96 µs/edge) — the O(files) signature. At 50,000 edges the two are within noise of each other because fixed per-open overhead (DataFusion context, ordinal-identity open, etc.) dominates at that size; the linear term only becomes visible once it outweighs the fixed cost, exactly as in the epic's own ladder data.

Peak resident memory (`VmHWM` after the open phase) at 400,000 edges: before 187.0 MB, after 161.6 MB — no regression; if anything a modest reduction, consistent with fewer full-file read buffers allocated during the removed verification passes. No new O(E) resident-memory term is introduced by this change; the bounded-memory property from #1094 is untouched (adjacency shard opens are unaffected).

I could not reproduce the exact multi-million-edge, sixteen-fold ladder rungs from #1388's own table within this session (that ladder is a separate, long-running, dedicated fixture); the numbers above are from an independently constructed fixture at a smaller but still meaningfully O(E)-dominated scale, intended to demonstrate the mechanism and its direction/magnitude, not to replace #1388's own ladder re-run (tracked as O6 in the design doc, owned separately).

## Out of scope (unchanged in this PR)

- O4 (XXH3 swap in `ordinal_identity_v4.rs`) — different agent, file untouched here.
- The administrative verify command (S7) — different agent; this PR's O3 tradeoff is exactly the work that command is meant to cover on demand.
- `supersession.rs`, `shaping_merge.rs`, `installation.rs`, `benches/`, per-phase I/O attribution, the CLI verify command — untouched, per the brief's file-ownership boundaries. Confirmed no open PR currently touches `materialization.rs` or `project_generation.rs` before starting.

Part of #1388

---
<a href="https://app.blacksmith.sh/CurateLabs/codesmith/graphforge/pr/1425"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792213138&installation_model_id=20486&pr_number=1425&repository=CurateLabs%2Fgraphforge&return_to=https%3A%2F%2Fgithub.com%2FCurateLabs%2Fgraphforge%2Fpull%2F1425&signature=5736731a30024f9ae09c6dfd23eb1fc149af2d7a18f531c1abf17532655e8dd2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


DecisionNerd:
## Blocking: this silently accepts corrupted Topology data

Converted to draft. An independent review with a working mutation test found that a **same-inode, same-length one-byte corruption of `topology/nodes.parquet` is accepted** — the open succeeds, and an ordinary query over the corrupted data succeeds and returns a count.

Reproduced against two worktrees at `cd964b69` and its parent `13632d4b`:

| commit | fresh `GraphForge::new` after flipping one byte in a hardlinked Topology payload |
|---|---|
| `13632d4b` (before this PR) | **refused** — `validation error: graph object digest does not match its address` |
| `cd964b69` (this PR) | **succeeded**, and `MATCH (n:Person) RETURN count(n)` then returned a row |

The refusal that disappears is `graph_files_inventory()`'s `verify_graph_object` call in the V2 loop, which O3 deletes.

### Why the identity check cannot cover this

`verify_hardlinked_materialization` checks `file_identity` (device + inode), declared length, and link count >= 2. Rewriting a file's bytes in place changes none of them — those are directory-entry properties, not content properties.

The cited precedent, `StableDirectory::link_child_into`, does genuinely compare `file_identity()` before and after linking; that half of the PR's claim is accurate. But identity proves "this entry points at the inode I just read", not "the bytes at that inode match the digest". Those coincide only if something hashed the bytes at a point you are relying on, and post-O3 nothing hashes Topology objects at all.

### Why lazy re-authentication on first read does not hold here

The commit message's central justification is that every real consumer re-authenticates lazily via `open_graph_object_by_digest` / `read_graph_object_by_digest`. Neither symbol appears anywhere in `crates/graphforge-exec/src` — verified by grep. Their callers are construction, catalog, GC, manifest-tree, ontology composition and export.

After `hydrate_graph_workspace` hardlinks the compact graph into a private workspace with route-encoded paths, the query engine opens those files **by path**. There is no digest lookup and therefore no "first real read" that re-authenticates. The justification is false for exactly the data class the test corrupts.

### Properties survive only by accident, and that is the #1392 pattern again

`property_overlay/inventory.rs` re-hashes every property route file against `entry.content_sha256` whenever `AuthenticatedPropertyInventory` is built, which happens on every ordinary open. It is untouched by this PR and wired only to Properties-role objects. So Properties stay protected by a wholly separate subsystem while Topology loses its only check.

That asymmetry also explains why `compact_graph_root_reopens_through_ordinary_api_and_rematerializes` still passes: its victim selection is `inventory.files.iter().find(|entry| entry.byte_length > 0)`, which lands on a Properties/Catalog entry in file order. It never exercises the O2 hardlink path. **That test is not evidence this PR is safe — it covers a path this PR does not touch.**

### The `ShardedCsrIndex::open` analogy does not transfer

`ShardedCsrIndex::open` is genuinely presence-only, but its row-read path validates an embedded per-shard checksum on decode — flip a byte and `row()` fails. Topology Parquet has no embedded self-check and no digest re-check on read.

### What is sound

**O1 is fine.** The `OnceLock` memoization is correctly per-instance and does not leak across opens. The measured win is real but lands on Topology's share of the bytes only, which is consistent with the ~2x rather than ~4x wall clock and the null result at 50K.

### To unblock

Either give Topology objects a real content check on first materialization — cheap is fine, but it must be a function of the bytes, not of the inode — or state explicitly and honestly that Topology data is unverified until `graphforge verify` runs. O3's comment already says something like that for the inventory sweep; O2 does not, and as written both are false in practice.

Whichever route, **ship the mutation test with it**. This PR currently has no test that would have caught this.

Note for #1414: its description documents that corrupting a referenced object is refused at ordinary open. That holds today and would remain true for Properties, but this PR makes it false for Topology.

DecisionNerd:
Superseded by #1435.

#1435 carries both commits — the O1 memoization from this PR (`4a92035b`, rebased) and the restoration of the content check this PR removed (`3e624070`) — so the optimization is not lost, only the data-integrity regression.

## Why this one is not landing

An independent review with a working mutation test showed that after this PR, a one-byte corruption of `topology/nodes.parquet` — same inode, same declared length — is **silently accepted at open, and an ordinary `MATCH (n) RETURN count(n)` then returns a count over the corrupted data**. Reproduced against `cd964b69` and its parent: the parent refuses with `graph object digest does not match its address`; this PR does not.

The justification in the commit message — that every real consumer re-authenticates lazily on first touch via `open_graph_object_by_digest` / `read_graph_object_by_digest` — is false for Topology-role objects. Neither symbol appears anywhere in `crates/graphforge-exec/src`. After `hydrate_graph_workspace` hardlinks the compact graph into a private workspace, the query engine opens those files **by path**, so there is no first read that re-authenticates.

Properties-role objects stayed covered only by accident, via `property_overlay::inventory`'s independent re-hash of every property route file — a mechanism this PR never touched and that does not reach Topology.

## What #1435 does differently

It restores `verify_graph_object` in `graph_files_inventory()`'s V2 loop, and because **O1's memoization is kept**, that check now runs **once per resolved generation instead of the 3-4 times per open** it ran before this PR. The redundancy that motivated this work is still removed; the guarantee is back. No durable-format change.

`entry.content_sha256` stays a full cryptographic digest because it is the content-addressed CAS lookup key, not merely a corruption checksum. A fast-checksum alternative belongs to #1417, which needs an inventory format change.

Ships with `hardlinked_topology_payload_corruption_is_refused`, which fails against this PR's head and passes with the restoration.

Part of #1388.

Issue #1416, opened 2026-09-17
fix(storage): shaped output bytes are not reproducible across sessions, because a wall clock is written into the runtime catalog


DecisionNerd:
## Summary

A wall-clock timestamp taken at session start is written into a **shaped output artifact**, so two ingests of byte-identical input in two different sessions produce different bytes. This is true on `main` today, independent of any parallelism work.

This matters now because the write-path redesign's determinism contract (#1387, step S5 range partitioning and step S6 concurrent shards) rests on the premise *identical logical input produces identical digests*. That premise is false today. Anyone proving a determinism property against `main` will either see a pre-existing failure and attribute it to their change, or scope their comparison to a single session and mistake a pass for proof.

Found by the agent on #1392 while removing redundant authentication passes. Verified independently in source before filing.

## Where it comes from

    crates/graphforge-storage/src/graph_construction.rs:1438
        session_now_micros: SystemTime::now()          <- wall clock, once per session

    crates/graphforge-storage/src/graph_construction.rs:565
        pub runtime_catalog_now_micros: i64            <- carried on the shape

    crates/graphforge-storage/src/graph_construction/shape.rs:379
        runtime_catalog_now_micros: self.checkpoint.session_now_micros

    crates/graphforge-storage/src/graph_construction/catalog.rs:158
        let output = "shaped-runtime-catalog.parquet"; <- written into a durable artifact

So the value reaches a file that ships as part of the shape output, and the file's bytes therefore differ per session.

## The part that makes this non-trivial: the timestamp is load-bearing

It is not decoration. Recovery uses it as a **session binding token**:

    crates/graphforge-storage/src/graph_construction/recovery.rs:275
        || shape.runtime_catalog_now_micros != checkpoint.session_now_micros

A completed shape whose recorded timestamp does not match the resuming checkpoint's is refused with `complete shape manifest inventory is incomplete`. That is a real correctness property: it stops recovery adopting a completed shape produced by a *different* session, which would silently pair outputs with the wrong inputs.

**So the naive fix is wrong.** Zeroing the field, or dropping it from the struct, removes the crash-recovery refusal along with the nondeterminism. Any change here has to keep the binding and remove only the nondeterministic bytes.

## Suggested shape of the fix (not prescriptive)

The binding and the artifact have different requirements, and today one value serves both:

- **Session binding** needs a value unique per session. It belongs in session state — the checkpoint and the shape manifest — where per-session variation is correct and expected.
[...]

PR #1429, opened 2026-09-17
perf(storage): materialize shape partitions concurrently (S6)

DecisionNerd:
## Summary

Stacked on #1387's S5 (`perf/1387-s5-range-partition`) — rebased onto S5's current head (`7e9f5eda`), which includes S5's Windows `#[cfg(unix)]` fix and its data-dependent partition-count cut (`min(partition_count, max(1, staged_identity_records / 16))`) landed after this branch was first opened.

S5 made partitions independent by construction but ran them sequentially. This PR makes the load-and-sort step of `FixedRangePartitioner::finish_optional` and `RowRangePartitioner::finish` concurrent:

- `load_partitions_ordered()` runs partition load+sort on a bounded worker pool (`std::thread::available_parallelism()`, capped by the number of sealed partitions — a scheduling decision, never a recorded format parameter), while still handing each sorted partition to the single writer thread in strict ascending partition-index order. Concatenation, digests and evidence totals are therefore unaffected by how many workers ran.
- **Memory bound**: the worker→writer channel is a rendezvous (capacity 0), so at most `threads` partitions are being actively computed plus at most `threads` completed-but-unconsumed results sit in a reorder buffer. Peak resident partition memory is bounded by `2 * threads * max_partition_bytes` — a constant multiple of thread count, never of partition count.
- `load_partition` on both partitioners no longer takes `&self`; each returns a freshly-zeroed `GraphConstructionEvidence` delta merged into the live evidence on the orchestrating thread via `merge_partition_load_evidence()` (every touched field is a sum or a max — order-independent, proven below across six forced thread counts).
- Cancellation stays at its pre-concurrency granularity: `load_partitions_ordered` keeps sole ownership of the caller's `cancelled` closure and hands `consume` a short-lived reborrow of it as a `&mut dyn FnMut() -> Result<(), GfError>` callback, so `write_bounded_row_groups` (per row group) and the fixed-width writer (per 4096 records) poll it exactly as often as before.

## Measured result: **1.34x on overall ingest wall-clock (~25% less time)** — real, but bounded, and it does not reach the product floor

Phase-time split inside `shape_canonical_inner` (wall-clock, one process, `524,288` / `8,388,608` edges — the ingest floor gate's own sweep):

| phase | 524,288 edges | share of shape | 8,388,608 edges | share of shape |
|---|---:|---:|---:|---:|
| sequential routing pass (untouched by this PR) | 0.928s | 17.5% | 17.227s | 34.6% |
| **concurrent finishes (this PR's region)** | 2.936s | 55.4% | 18.072s | 36.3% |
| unaccounted (surrogate assignment, validation, `resolve_endpoint_surrogates`'s own routing loop — also untouched) | 1.441s | 27.2% | 14.493s | 29.1% |
| **shape total** | 5.305s | 100% | 49.792s | 100% |

At 16x more edges, the untouched sequential routing pass grew **18.6x** (super-linear — it is getting relatively worse), while this PR's concurrent region grew only **6.2x** (sub-linear: roughly **2.6x effective parallelism** out of 16 threads, a real but partial win, not close to 16x). Working from that 2.6x and the official floor-gate run's absolute wall time (8,388,608 edges, 84.552s total, 49.792s of it shape): making the concurrent region fully sequential would cost ~46.9s instead of 18.072s, putting shape at ~78.7s and total ingest at ~113.4s instead of 84.6s — **S6 is worth 1.34x on overall ingest wall-clock**. That also bounds the remaining headroom in this region: driving it to zero from here buys only a further 1.27x. The larger lever is the untouched 63.7% of shape (routing + the unaccounted slice) — filed as **#1437**, not this PR's task.

**This does not reach the 1,000,000 edges/s product floor, and was never going to on its own.** Measured on the official run: 66,567 edges/s and 50,716 edges/s at the two swept sizes — 15.0x and 19.7x short of the floor. CPU per edge (13.61 / 11.77 µs) also misses the <9.0 µs target. Effective cores measured 0.91 → 0.60 (falling with scale) on that run.

**Reproducibility caveat — read the phase fractions, not the absolute numbers.** Two more runs of the *identical* binary on this shared host gave 79,259/99,212 edges/s (cores 0.88→0.96, **rising** with scale) and 64,433/93,459 edges/s (cores 0.85→0.95) — the opposite trend from the official run, on the same code. Absolute wall-clock throughput and effective-cores are not a stable signal on this host; the phase-time *fractions* above (measured within one process, so far less sensitive to external contention) are what this PR leads with.

**Peak RSS**: 899,688 KB (≈879 MiB) against the 4 GiB budget — 21.5%, comfortable headroom, consistent with the `2 * threads * max_partition_bytes` bound.

**The `ingest floor gate: pass` line in the log means the 15,000 edges/s contention-tolerant CI regression guard passed, not the 1,000,000 edges/s product floor.** Do not read one for the other.

## Determinism: unaffected by concurrency

- The four fixed-width digests are byte-identical to S5's (and to this branch's own pre-rebase measurement): `shaped-identities.run` `6ea5b8462fc1b5d1…`, `shaped-node-details.run` `d7eac2fdc3d4e0e3…`, `shaped-edge-details.run` `4b9d9dd1f920b533…`, `shaped-edge-endpoints.run` `0b3c8415df3d42d6…`.
- `unpinned_sessions_differ_only_in_the_wall_clock_runtime_catalog`: 7 artifacts compared, exactly one differs (the pre-existing #1416 wall-clock catalog) — concurrency contributes no second difference.
- New test `concurrent_worker_counts_reproduce_the_sequential_loader`: forces worker counts 1/2/3/5/8/16 (partition_count=256, so partitions stay well above thread count) with a deterministic completion-order jitter enabled above threads=1, and asserts every fingerprint and every partition-layout evidence field matches the threads=1 run (which takes `load_partitions_ordered`'s sequential fallback branch verbatim) exactly.
- `cargo test -p graphforge-storage --lib graph_construction::tests:: -- --nocapture --test-threads=1`: **53 passed, 1 failed** in 219.65s. The one failure, `range_partition_work_is_linear_across_production_chunk_counts`, is S5's own in-flight fixture-growth fix for its new data-dependent partition cut (not this PR's code, not touched here).
- `cargo test -p graphforge-storage --lib graph_construction::partition::tests`: **12 passed, 0 failed** (a sibling module to `graph_construction::tests`, not covered by the filter above).
- `cargo clippy -p graphforge-storage -- -D warnings` (matches CI's `cargo clippy --workspace -- -D warnings` in `.github/workflows/test.yml`, scoped to the one crate touched): clean.

## Not this PR's task

- **#1437** (filed by the coordinator during this review): the sequential routing pass is the larger, growing lever — 18.6x wall-clock growth for 16x data, climbing from 17.5% to 34.6% of shape's own time in one scale step. That is the next step toward the 1,000,000 edges/s floor, not this one.
- Whether the routing pass and the per-family finish could overlap (pipelining routing of family N+1 while finishing family N) is unexplored here.

Part of #1387

> [!NOTE]
> ### Replace merge-tree shaping with concurrent range partitioning
> - Shaping now samples staged identity keys to build deterministic UUID range splitters, routes each staged record into its partition, materializes partitions independently, and concatenates sorted partitions in order instead of using the external merge tree.
> - Adds `PartitionPlan`, `IdentitySampler`, and `PartitionBalance` types in the new [partition.rs](https://github.com/CurateLabs/graphforge/pull/1429/files#diff-10a521f6188ac38ae81bd401f78df9a82ca88faf2c3bf676288de540bd82e4cb); `GraphConstructionBudgets.partition_count` defaults to 256 (max 4096) and splitters are recorded in `ShapeIntent`.
> - Removes the `shaping_merge` module and all `merge-` artifact names; recovery and inventory checks now use `staged-`, `shaped-`, and `part-` names via the shared `is_shape_scoped_name` predicate.
> - `PartitionBalance.assert_balanced` rejects identity partition skew above 4x the mean when the minimum-row threshold is met.
> - Behavioral Change: permanent Parquet files now write a fixed `CREATED_BY` footer instead of inheriting the Arrow version, changing file bytes and digests; shape artifacts are renamed from `merge-*` to `part-*`/`staged-*`/`shaped-*`, so older sessions or out-of-tree readers expecting merge-tree artifacts are incompatible.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 57d3ea4.</sup>
> 
>


DecisionNerd:
## Hold: this must not land before the endpoints skew is fixed

Flagging a constraint found while diagnosing the S22 RSS refusal, because it makes this PR unsafe to land in its current form and the reason is not visible from the diff.

**S6's memory bound is `2 x threads x max_partition_bytes`.** That is fine when partitions are balanced. They are not.

`shape.rs:358` carries exactly one production balance assertion:

```rust
identities.balance().assert_balanced("staged identity")?;
```

Its own comment calls it *"the only check that can tell a working range partition from a catastrophically skewed one."* **Endpoints, node details and node rows are unchecked.**

Endpoints are keyed by **node** UUID but routed with **identity** splitters sampled from a joint domain that is roughly 94% edges. Nodes and edges occupy disjoint UUID bands, so all 2E endpoint records land in about 16 partitions out of 256. Verified arithmetic: at S19 that is ~1.05M records per partition, a `Vec` doubling to 2^21, and `2^21 x ENDPOINT_WIDTH(33) = 69.21 MB` — matching a heaptrack S19-minus-S18 diff at `FixedRangePartitioner::load_partition` exactly.

**The consequence for this PR:** the endpoints partition at S22 is projected at ~554 MB. Multiplied by S6's bound at 16 threads that is roughly **17 GB of resident memory** against a 4 GiB budget. S6 would turn a skew that currently inflates RSS into one that exhausts it.

After the skew is fixed, the largest family becomes edge details at roughly 80 MB per partition at S22, giving ~1.3 GB at 16 workers — acceptable, but only with the worker count capped.

## What this PR needs before it can land

1. **Rebase after the per-family splitter fix lands** (tracked on #1439). Not before.
2. **Cap workers at 8, not `available_parallelism()`.** The host is an AMD Ryzen 7 3800X — **8 physical cores, 16 threads**. The measured work is SMT-hostile (SHA-NI, zstd, memcpy, sort at IPC 2.08), so siblings add scheduling pressure and memory footprint without adding throughput. 16 workers doubles the memory bound for little gain.
3. **Re-verify the four fixed-width digests and the determinism suite** after the rebase, since the splitter change alters partition boundaries for three families.

## Unrelated correction worth having

The plan previously recorded this PR as "worth 1.34x." That figure has no same-fixture baseline. The clean ladder measures S5's ingest at **1.30x** over the pre-redesign engine at both S18 and S20 — and that is S5 plus seven other merged commits, not S6. **S6's own increment at rung scale is unmeasured.** Worth establishing before the PR body claims a number.

Nothing here is a criticism of the implementation, which verified byte-identical output across six thread counts with jitter. The problem is underneath it.

DecisionNerd:
## The stated memory bound does not hold: the reorder buffer has no backpressure

Reasoned from source on `perf/1387-s6-concurrent-shards`, **not yet demonstrated by a test** -- the test is proposed at the end. Flagging it before the rebase because it changes what the rebase should do.

### What the module doc claims

```
//! **Memory bound.** The channel between workers and the orchestrating thread
//! is a rendezvous (capacity 0), so a worker's send blocks until the
//! orchestrator receives it. A result that arrives out of turn waits in a
//! reorder buffer; because there are only `threads` workers, at most
//! `threads` results can be ahead of the one the orchestrator is waiting for.
```

### Why that inference does not follow

The orchestrator receives **eagerly and unconditionally**:

```rust
for (index, result) in &rx {
    ...
    pending_values.insert(index, value);
    while let Some(want) = next_wanted {
        let (Some(value), Some(delta)) = (pending_values.remove(&want), pending_deltas.remove(&want))
        else { break; };
        ...consume(want, value, ...)?;
        next_wanted = wanted.next();
    }
}
```

There is no condition gating the receive on `pending_values.len()`. A `sync_channel(0)` send unblocks as soon as the receiver **takes** the value -- not when it finishes processing it. So every iteration of the `for` loop releases one worker, which immediately does `next.fetch_add` and starts another partition.

The argument conflates **workers in flight** (correctly bounded by `threads`) with **completed results accumulated** (not bounded at all).

### The failure sequence

1. Worker A takes partition 0. It is the slow one.
2. Workers B..P take 1..15, finish fast, send. The orchestrator receives all 15 and buffers all 15 -- it cannot consume any, because it is waiting for 0.
3. Those workers are now unblocked, grab 16..30, finish, send, buffer.
4. Repeat until the item list is exhausted.

`pending_values` grows until partition 0 completes. **Worst case it holds `N - 1` partitions** -- bounded by partition count, which is exactly what the doc says it is "never" bounded by.

### Why this is not a theoretical worst case here

This is the skew scenario precisely. With the endpoints skew, one partition is enormous and slow while the rest are small and fast -- the exact shape that maximises the buffer. **So the skew does not just make one partition large; it makes the reorder buffer hold every other partition while waiting for that one.** Two multipliers compounding, and the second one was believed to be bounded.

It also compounds with #1441 (detail records held padded, a 5-6x in-memory multiplier), since what accumulates in the buffer is those same padded `Vec<[u8; N]>` values.

### On host-tuning, since it came up

`shaping_worker_count` is fully host-derived and contains no machine constants:

```rust
std::thread::available_parallelism()
    .map_or(1, std::num::NonZeroUsize::get)
    .min(work_items.max(1))
```

So this is **not** over-fitted to the 8c/16t bench host. The problem is the opposite: the designed bound `2 x threads x max_partition_bytes` *scales with core count and is never capped*. This host's 16 threads is where it first hurts; a 128-core machine would be 8x worse. Under-constrained for every host, not tuned to one.

(Note `available_parallelism` does honour cgroup quotas and affinity, so under BenchExec it returns the restricted count -- which means the bound moves with the harness configuration too.)

### What the rebase should carry

1. **Real backpressure.** Gate the receive on buffer occupancy, or size the channel so workers block once `threads` results are outstanding. Then the doc's bound becomes true rather than asserted.
2. **Cap workers at physical cores, not SMT threads.** These are memory-bandwidth-bound sorts; the second hyperthread on a core adds buffer occupancy without adding throughput.
3. **Correct the module doc.** It is currently load-bearing -- it is the reason nobody looked here.

### The test that would settle it

Use the existing `test_support::set_worker_count_override` plus `shaping_worker_jitter`, but invert the stagger so **partition 0 is the slowest** rather than index-proportional. Assert peak `pending_values.len()` stays `<= threads`. Today that assertion should fail; after backpressure it should pass. The knobs for this already exist in the branch.

PR #1444, opened 2026-09-17
perf(storage): stop re-reading encoded artifacts after the encoder writes them

DecisionNerd:
## What this changes

The encoder used to finish by re-reading every artifact it had just written and
re-computing its SHA-256 against the inventory it had just installed
(`authenticate_inventory` at the end of `encode`). That pass is now the
control-only `authenticate_inventory_control`: the inventory's structural
invariants and its retained-parent references, with no payload bytes read.

Part of #1387, in the direction of #1384 ("hash once on write"). This is the
`encode_write_postwrite_authentication` lever; the recovery-phase reclaim
passes are **not** touched here and are reported separately.

## What the pass protected, and what protects it now

Every artifact digest in the inventory is computed by the single pass that
writes the bytes (`CountingWriter` wraps `DurableFileCacheWriter`, so the
digest sees exactly the bytes that reached the file). The post-write re-read
therefore compared bytes in this process's own page cache against a digest this
process computed a moment earlier. The only disagreement it can observe is a
writer-side defect or a same-identity concurrent mutation, which ADR 0013's
threat model excludes.

The boundary that **consumes** the encoded artifacts is the CAS install at
publication, `install_graph_object_file_with_lease`, which copies and hashes
every artifact against the inventory and refuses a mismatch
(`graph object source digest or length changed during install`). It ran
before this change and it runs unchanged after it. A corrupted encoded artifact
is refused at first touch; an eager sweep in the encoder adds no refusal the
install does not already make.

Mechanisms touched, by name:

- `graph_construction_encoding::encode`, post-write pass: **demoted** to
  control-only. Payload re-read removed.
- `authenticate_inventory_payloads`: **unchanged**, still called by the reclaim
  pass in `supersession.rs` and by the encoder's resume path.
- `install_graph_object_file_with_lease`: **unchanged**; now the sole owner of
  encoded-payload corruption refusal on the single-process path.
- #1269 (`authenticate_shaped_output`, incomplete-shape recovery): **not
  touched**. Different artifact set, different boundary.

## Mutation proof

`publication_refuses_same_inode_encoded_payload_corruption_at_cas_install`
flips the first byte of `topology/surrogate_tails.parquet` in place after
`encode_canonical`, asserts inode and length are preserved, then calls
`publish_canonical` directly — which never runs the reclaim sweep — and asserts
the refusal **by the CAS install's own message**, not on the end-to-end result.
It also pins that `authenticate_inventory_payloads` refuses the same bytes with
its own message, so the two refusals are visibly independent.

The removed pass could never have caught this mutation (it ran before the
mutation), so this is not a "fails on unfixed code" test; it is the #1392
pattern's second half — proof that the consuming boundary does not depend on
the deleted sweep. The probe-by-removal in the other direction is the storage
suite: no existing test depended on the post-write re-read.

## Measured

Real `gf` binary, the ladder's five-process ingest (`begin`, `register-parquet`
x2, `validate`, `commit`), Graph500 generator inputs, per-phase receipts from
the `commit` step. Byte counts are deterministic; this is not a timing claim.

| | S14 (262,144 edges) | S16 (1,048,576 edges) |
|---|---:|---:|
| encode phase read bytes, before | 65,869,006 | 264,941,854 |
| encode phase read bytes, after | **56,898,946** | **228,064,468** |
| removed | 8,970,060 (= encoded graph, 1.0x) | 36,877,386 (= encoded graph, 1.0x) |
| total application read bytes per edge, before | 1,191.4 | 1,195.9 |
| total application read bytes per edge, after | **1,157.2** | **1,160.8** |
| | −2.9% | −2.9% |
| encode phase write bytes | unchanged (10,255,220 / 42,254,776) | |
| recovery_reauthentication | unchanged (35,887,923 / 147,520,407) | |

Projected at S20 from the `clean-9269362f` receipt: the encode phase drops by
one full read of the 678 MB encoded graph, 4.40 GB → ~3.72 GB; total logical
reads 20.77 GB → ~20.09 GB (−3.3%). The remaining 3.7 GB in that phase is the
encoder consuming its shape inputs (fixed-width reads ~63%, source spool copy
~30%), not authentication.

## Verification

- `make pre-push-fast`
- `cargo test -p graphforge-storage --lib` (1158 passed), plus, run
  separately with non-zero counts: `graph_construction::tests::` (55),
  `graph_construction::tests::lifecycle_budget` (22), `recovery::tests` (34).
- `cargo test -p graphforge-api`
- Determinism: the contract is self-consistency (identical logical input,
  identical shaped and encoded bytes across sessions, project directories and
  partition counts), not equality to a recorded constant. Run as
  `graph_construction::tests::determinism` (8 passed) and
  `graph_construction::partition::tests` (12 passed), filtered separately
  because they are sibling modules. Nothing here touches the shape or the
  encoded wire format.

---
<a href="https://app.blacksmith.sh/CurateLabs/codesmith/graphforge/pr/1444"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792260074&installation_model_id=20486&pr_number=1444&repository=CurateLabs%2Fgraphforge&return_to=https%3A%2F%2Fgithub.com%2FCurateLabs%2Fgraphforge%2Fpull%2F1444&signature=08ac959bf13e83cb4917d7f2aae3f66178005ca3b9785be24f1c9ed50de1826d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


DecisionNerd:
## Merge ordering with PR #1450

This PR and PR #1450 both modify `crates/graphforge-storage/src/graph_construction/encoding_publication/tests.rs`.

They need to merge in order rather than race — whichever is second should rebase and re-run, not resolve blind. Both are re-authentication reductions under #1384, so the second one's test expectations depend on what the first one left in place.

Issue #1433, opened 2026-09-17
test(g500-ladder): the phase metric policy table assumes work proportional to data, which the redesign removes


DecisionNerd:
## Summary

`PHASE_METRIC_POLICIES` in `crates/graphforge-api/tests/scale_g500_ladder.rs` encodes the **pre-redesign architecture**, in which nearly every lifecycle phase did work proportional to the data. The write-path redesign exists to stop doing proportional work. So as it lands, rows in this table fail one by one — and each failure needs a judgement that cannot be made generically.

Four pull requests hit this in one night, **each on a different row**:

| PR | row and field | observed 1x/2x/4x | verdict |
|---|---|---|---|
| #1412 | `recovery_reauthentication.read_calls` | 69, 68, 70 | legitimate regime change — **fixed**, see below |
| #1413 | `cas_install_read_write.write_bytes` | no positive growth | install-by-rename makes writes ~O(1) |
| #1425 | `hydration_verification.read_bytes` | 87475, 87479, 87479 | **a real defect — see the warning below** |
| #1430 | `fsync_synchronization.fsync_calls` | 2195, 2089, 2063 | merge tree deleted, fewer files to fsync |

Verified by checking out each branch and running the failing test, not by assuming a shared cause.

## The warning that makes this urgent

**A flat metric is not automatically a regime change. On #1425 it was a genuine data-integrity bug.**

#1425's `hydration_verification.read_bytes` went flat because the PR removed the only content verification of Topology-role graph objects on the open path. An independent review proved a one-byte corruption of `topology/nodes.parquet` — same inode, same length — is silently accepted, and a query then returns a count over corrupted data. **The table's assertion caught that.**

So the reflex "the redesign removes proportional work, therefore flat is expected" is true often enough to be dangerous. It is exactly the argument that would have waved #1425 through. Every row needs a judgement, and the judgement has to distinguish *stopped doing redundant work* from *stopped doing necessary work* — which look identical at a growth assertion.

## What #1412 established, as a worked example

[...]

PR #1450, opened 2026-09-17
perf(storage): stop re-hashing the encoded inventory three times per ingest

DecisionNerd:
## Summary

`recovery_reauthentication` was **exactly 4x the encoded graph** on every successful ingest, from four sites. Three of them re-hash bytes already hashed earlier in the same process. This takes those three to **zero payload bytes read**, without removing the refusal.

| `recovery_reauthentication` read_bytes | S14 (262,144 edges) | S16 (1,048,576 edges) |
|---|---:|---:|
| `validate` receipt, before | 8,970,060 | 36,877,386 |
| `validate` receipt, after | **0** | **0** |
| `commit` receipt, before | 35,887,923 (136.9 B/edge) | 147,520,407 (140.7 B/edge) |
| `commit` receipt, after | **8,977,743 (34.2 B/edge)** | **36,888,249 (35.2 B/edge)** |
| removed | 26,910,180 = 3.0x encoded graph | 110,632,158 = 3.0x encoded graph |
| **total application read bytes/edge** | 1,191.4 -> **1,088.7** (&minus;8.6%) | 1,195.9 -> **1,090.4** (&minus;8.8%) |

The remaining 1x is site 4 (`authenticate_public_successor`), untouched here.

Measured on the real `gf` through the ladder's five-process ingest. Byte counts are deterministic; no timings are claimed (the host was loaded).

## The change

The encoded-successor branch of `reclaim_superseded_payloads_cancellable` now calls `authenticate_encoded_artifact_identities` instead of `authenticate_inventory_payloads`: for every inventory artifact, open the file and require its identity key (`{volume:016x}:{file_id}`) to be present in the checkpoint's `storage_active_identity_allocated_bytes` ledger with the same allocated bytes, link count == 1, and logical length == inventory bytes. **Zero payload bytes read.**

**Why not #1392's precedent exactly.** The shape branch checks identity against `ArtifactReceipt.identity`, but `ConstructionEncodedArtifact` records only `path`/`bytes`/`sha256` — no identity. Adding one would change `inventory_authority_sha256`, a durable format change. The ledger used instead is written by `record_encoded_active_artifacts` in the **same checkpoint write** that pins `encoding_inventory_sha256`, so it is persisted before any reclaim can run in the commit process and needs no format change.

## Untouched, by name

`revalidate_named()`, `project_identity`/`session_identity` matches, the inventory authority check, `authenticate_retained_successors`, the receipt-chain walk, `retire_payload`, **`authenticate_shaped_output` (#1269 — the correctness floor)**, `authenticate_inventory_payloads` itself (still used by the encoder's reuse path), and `install_graph_object_file_with_lease`. No change to `graph_construction_encoding.rs`, so this should merge cleanly alongside #1444.

## Proof

Every refusal is asserted **at the boundary that owns it**, by that boundary's message:

- `ladder_path_refuses_same_inode_encoded_corruption_at_cas_install` — flip a byte in place (inode and length asserted unchanged), reopen, assert `prepare_canonical_encoding` charges **exactly zero** recovery bytes, then `publish_canonical` refuses with `graph object source digest or length changed during install`. **This test fails on unfixed code.**
- `reclaim_refuses_encoded_artifact_inode_replacement` — byte-identical copy under a new inode, refused at reclaim.
- `reclaim_refuses_encoded_artifact_length_and_link_changes` — same-inode append and an extra hard link, both refused.

Three existing tests retargeted to the new owning boundary; none deleted or loosened.

## The authorised trade

Reclaim retires staged predecessors before publication refuses. A corrupted encoded artifact is **still refused fail-closed at the CAS install**, `CURRENT` is untouched and the session is not Published — but it can no longer be re-encoded from its predecessors, so the import must be re-run from source. This is **recoverability-without-restart**, accepted deliberately; `construction_lifecycle_tests.rs` now states it explicitly.

## Policy row (#1433): cheaper, not absent

`recovery_reauthentication.read_calls` went 68-72 -> 54-56, tripping a floor of 55. The drop is **exactly 14** — the re-read this PR authorises removing — with the other two contributors (successor auth ~27-31, drill 27) intact. `read_bytes` still passes `SCALE`.

The row was **not relaxed**: `RECOVERY_REAUTHENTICATION_MIN_ABSOLUTE_CALLS` is re-derived at **45** from the two remaining contributors (losing either leaves <=~29 against an observed min of 54), with the attribution recorded in the comment. The synthetic fixture pins 60 so a uniform-drop mutation still exercises the floor.

## Verification

- `graphforge-storage --lib`: 1160 passed. Determinism 8, lifecycle_budget 22, partition::tests 12, recovery::tests green (filters run **separately**).
- `graphforge-api --test bdd`: 118 passed; **openCypher TCK 3897/3897**, zero regressions.
- `graphforge-api --test scale_g500_ladder`: 30 passed.
- `cargo clippy --workspace -- -D warnings`: clean. `make pre-push-fast`: passed.

Part of #1384. Related #1417 (whose acceptance criterion was corrected on the issue), #1433.

---
<a href="https://app.blacksmith.sh/CurateLabs/codesmith/graphforge/pr/1450"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792263860&installation_model_id=20486&pr_number=1450&repository=CurateLabs%2Fgraphforge&return_to=https%3A%2F%2Fgithub.com%2FCurateLabs%2Fgraphforge%2Fpull%2F1450&signature=640776b432b46d893ef74d4082132a7fe8663a62c669e23c27cc7be5c7d3760f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


DecisionNerd:
## Merge ordering with PR #1450

This PR and PR #1450 both modify `crates/graphforge-storage/src/graph_construction/encoding_publication/tests.rs`.

They need to merge in order rather than race — whichever is second should rebase and re-run, not resolve blind. Both are re-authentication reductions under #1384, so the second one's test expectations depend on what the first one left in place.

PR #1491, opened 2026-09-19
fix(storage): bound contiguous partition routing buffers

### LINKED ISSUE #1445: perf(storage): the PartitionRun accumulator is unbounded and grows with partition size
## Summary

PR #1440 cleared the S22 `rss_bounded_or_plateaued` gate and cost **~23% ingest throughput**. Recovering it is worth **~2.40 h at S26** — more than the entire nine-item constant-factor plan combined, and the single largest item on the board.

| | S18 | S19 | S20 |
|---|---:|---:|---:|
| before #1440 | 99,921 | 102,516 | 101,226 edges/s |
| after #1440 | 80,408 | 78,171 | 77,412 edges/s |
| | -19.5% | -23.7% | **-23.5%** |

S20 wall 319s -> 381s. Projected S26 moves 5.53 h -> 7.93 h against a 3.20 h ceiling.

**#1440 must not be reverted.** It takes projected S26 ingest RSS from 5.57 GB to 2.17 GB against a 3.44 GB limit (slope 5.08 -> 1.94 B/edge). The memory fix and the throughput loss are believed separable.

## Mechanism

#1440 removed the per-partition `BufWriter::with_capacity(SPILL_BLOCK_BYTES /* 64 KiB */, hashing)` and replaced it with `PartitionRun` (`graph_construction/shape.rs:1056`), which accumulates a whole contiguous same-partition run into a `Vec<u8>` and writes it once. It flushes **only on partition change**.

The routing key is the record's leading 16 bytes, which is also the staged run's sort key, and the range partition is monotone — so runs really are contiguous. (An earlier "scattered writes" hypothesis was checked and killed.)

Which means: with ~256 partitions and ~33.5M endpoint records at S20, a run averages **~131k records ≈ 4.3 MB**. The write buffer went from a fixed 64 KiB (L2-resident) to a multi-megabyte `Vec` that grows by reallocation, and `clear()` retains capacity so it stays at peak.

**This predicts both symptoms at once**: throughput down ~23%, and ingest RSS still growing ~1.94 B/edge after the per-partition buffers were removed.

## The fix

Flush `PartitionRun` when `bytes.len()` crosses a bound **as well as** on partition change. Writes stay large enough to amortize the syscall, the buffer stays cache-resident, and the accumulator stops scaling with partition size.

Same shape as the #1442 decision: **reprice a bounded constant, do not remove the mechanism.** #1440's error was deleting the buffer outright when a bounded one was what it needed.

## Acceptance — BOTH, or it is not a fix

- **Throughput back to ~100k edges/s** at S18/S19/S20.
- **Ingest RSS still ~91/97/113 MB** with slope near 1.9 B/edge. The memory win is what unblocks S22 and must not regress.
- Report the bound curve (64 KiB / 256 KiB / 1 MiB), not a single value chosen by taste.
- Durable artifacts self-consistent via the determinism suite. TCK green.

## Verify first

Histogram `PartitionRun.bytes.len()` at every flush, per family, at S18. If mean flush size is in the megabytes, confirmed. **If flushes are already small this hypothesis is dead** and the cost is in `DurableFileCacheWriter`'s per-write window arithmetic instead.

Part of #1387. Follows #1439 / PR #1440.

DecisionNerd:
Bounds `PartitionRun` at 64 KiB, flushing on byte capacity or partition change while preserving whole records, row counts and output order. Oversized individual records bypass the accumulator.

Three clean rotating-order S18 comparisons selected 64 KiB: median complete-ingest wall was 28.798 s against the unbounded baseline's 29.320 s. The selected candidate passes the recorded S18/S19/S20 throughput floors; contended observations are retained and excluded. The checked-in report and JSON contain the protocol, results, limitations and binary/input hashes. This establishes a bounded buffer without a material regression under the predeclared selection rule, not a throughput-improvement claim.

Validation:
- Storage suite: 1,202 passed, six existing ignored tests; final-bound regressions also rerun.
- API BDD: 118 scenarios passed; openCypher: 3,897 scenarios passed, zero regressions.
- `make pre-push-fast`, gate-registry validation and formatting passed.
- CodeRabbit reviewed record framing, output equivalence and durability interactions with no blocking findings: https://github.com/CurateLabs/graphforge/pull/1491#issuecomment-5743246674.

Closes #1445

> [!NOTE]
> ### Bound contiguous partition routing buffers in `PartitionRun` to 64 KiB
> - Adds a byte bound to the `PartitionRun` accumulator in [shape.rs](https://github.com/CurateLabs/graphforge/pull/1491/files#diff-ad02ba3d3f61e50e6b6151e51f788bfa06dca9041198b91cf32055bc50125629), defaulting to 64 KiB in production and configurable for tests.
> - The `push` path flushes partial data before the buffer would exceed the bound and routes an individual record at or above the bound directly instead of buffering it; records are never split.
> - Adds tests in [tests.rs](https://github.com/CurateLabs/graphforge/pull/1491/files#diff-09d2f33550ded61b734bab3d1bab7ae3ec8c1987460117eb55920244872b6486) covering fixed-width and compact records across multiple bound sizes, asserting byte-for-byte output equivalence with the unbounded path.
> - Adds benchmark evidence artifacts under [docs/development/evidence/](https://github.com/CurateLabs/graphforge/pull/1491/files#diff-69c2e9b65c65c3d1bc09deb95219e235cd276b00850c682996d359c7b5ab6252) documenting bound selection and correctness checks.
> - Risk: `PartitionRun` now retains at most the configured bound of buffered wire bytes; any caller assuming full-run buffering before flush will see earlier, partial flushes. The new bound is enforced in [shape.rs](https://github.com/CurateLabs/graphforge/pull/1491/files#diff-ad02ba3d3f61e50e6b6151e51f788bfa06dca9041198b91cf32055bc50125629) `push`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7eb882f.</sup>
> 
>


DecisionNerd:
@coderabbitai review

Please focus on whether byte-bound flushes preserve record framing, partition transitions, row accounting and durable output, and whether the regression tests cover those contracts. The 1 MiB constant is provisional pending the requested 64 KiB / 256 KiB / 1 MiB quiet-host curve; no performance claim is ready yet.

